PR #443, opened 2014-11-15
Loader uses @include instead of require

dg:


TomasVotruba:
New stable 4.0 is here, so 2.8 is not officially supported anymore.

dg:
Why you didn't merge it?

GrahamCampbell:
Because 2.8 is not supported anymore.

dg:
This commit is here for two months…

GrahamCampbell:
I assume it required review by @kukulich, and there was no response, so then this pull request remained open beyond the time period where 2.8 was maintained, so is now closed.

dg:
I understand, but obvious things can be fixed without review by one of authors.

TomasVotruba:
As @GrahamCampbell said, @kukulich didn't merge it in last 2 months. I was waiting till stable release with this.

I take responsibility only for v4+ since I don't understand all consequences in v2.8.

PR #444, opened 2014-11-16
'mixed' type is not the same as 'undefined' vol. 2

dg:
and Deprecated methods are marked with class 'deprecated'


TomasVotruba:
New stable 4.0 is here, so 2.8 is not officially supported anymore.

dg:
It means that it is in 4.0 fixed or won't fix?

TomasVotruba:
You send PR to master in #445, so it should be fixed.

Issue #215, opened 2013-01-04
Make collapse method rows after open


doapp-ryanp:
As a future enhancement could you please make the method rows collapse-able after you expand them?

The methods summary rows that I am talking about are this element:

`<table class="summary table table-bordered table-striped" id="methods">..`

When I click on a row in that table, it expands, but then I can not collapse it back again.

PR #503, opened 2015-01-02
Added expand/collapse toggle to descriptions

crackertastic:
I've always liked ApiGen since the first time I used it (v2.8) and the only one whine I have ever really had about it was the inability to collapse a description for methods/properties after expanding them.  

A particular page of the documentation can get crowded fairly quickly as multiple rows are expanded and in my opinion the readability suffers because they can not be collapsed afterwards.

After reviewing previous issues I found that this has been brought up in the past in issue #215 which has been closed because it had gone stale and there was no PR for it.  After digging through some of the theme files I found a block of code in `main.js`  that has the code to expand those descriptions.  I've updated it to add a toggling effect.  I made the change on both the default and bootstrap themes.

Hope it is useful and can be merged in with the next release.

### DIFF
--- a/src/templates/bootstrap/js/main.js
+++ b/src/templates/bootstrap/js/main.js
@@ -130,11 +130,19 @@ $(window).load(function() {
 
 	// Open details
 	if (ApiGen.config.options.elementDetailsCollapsed) {
+		var trCollapsed = true;
 		$('tr', $content).filter(':has(.detailed)')
 			.click(function() {
 				var $this = $(this);
-				$('.short', $this).hide();
-				$('.detailed', $this).show();
+				if (trCollapsed) {
+					$('.short', $this).hide();
+					$('.detailed', $this).show();
+					trCollapsed = false;
+				} else {					
+					$('.short', $this).show();
+					$('.detailed', $this).hide();
+					trCollapsed = true;
+				}
 			});
 	}
 
--- a/src/templates/default/js/main.js
+++ b/src/templates/default/js/main.js
@@ -130,11 +130,19 @@ $(window).load(function() {
 
 	// Open details
 	if (ApiGen.config.options.elementDetailsCollapsed) {
+		var trCollapsed = true;
 		$('tr', $content).filter(':has(.detailed)')
 			.click(function() {
 				var $this = $(this);
-				$('.short', $this).hide();
-				$('.detailed', $this).show();
+				if (trCollapsed) {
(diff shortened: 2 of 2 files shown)


webmozart:
I'd would prefer to have the method descriptions in a different place than the summary. Currently, the summary is a bit cluttered:

apigen-methods

I think the Javadoc renderer does this better by separating the summary from the details:

Summary: 

javadoc-summary

Details:

javadoc-details

IMO the links of the method names in the summary should link to the details. I was quite confused when I clicked on the method name and was brought to the source view instead of the detail view for the first time (and some more times..). The source should be referenced by a "source" link in the details view IMO.

crackertastic:
On the method summaries/descriptions, I don't fully agree.  Whereas I do appreciate the separation the Javadoc implementation does (which I do like), I also enjoy the condensing that is done by ApiGen.  My beef with ApiGen is that once you open up a summary, you can't close it without refreshing the page.  Perhaps judicious use of white space or separators would help un-clutter the summary.  Of course, this is largely a matter of preference.  Maybe an optional switch could be made that would allow a user to generate their code in either fashion.

Now on the linking to code...you are spot on.  There isn't much of a reason I can think of that the method names link to the code document.  As you said they should link to a description (if they are separate) or toggle the description visibility.  Another link could be nearby that would allow the user to navigate to the source code.

webmozart:
A toggle in the configuration to switch between inline and separate method details would be lovely. :)

TomasVotruba:
@crackertastic Sorry for delayed reply. I checked the PR and it seems simple and direct.

Could I ask you to send PR to particular repos with templates?

- https://github.com/ApiGen/ThemeDefault
- https://github.com/ApiGen/ThemeBootstrap 

I shall merge then. Thank you

TomasVotruba:
@webmozart Seems like a good idea, I find it confusing sometimes too. Could you open new issue/PR in default theme repository?

Also links to docs that is on screenshots would be helpful. Thank you.

crackertastic:
@TomasVotruba Sure thing. I will put together PRs for both of those repositories in the near future. Thanks for considering it.

crackertastic:
@TomasVotruba I've sent the PR to the default theme repo, looks like this already handled in the bootstrap theme (which looks to have been upgraded), so I didn't make any changes there.

TomasVotruba:
Perfect, thanks!

PR #534, opened 2015-01-28
Support static magic method

ct-jensschulze:
Added support for static magic methods and correct type hinting:

/**
 * @method myMethod(MyClass $name, MyClass $slug)
 * @method static myStaticMethod(MyClass $name, MyClass $slug)
 * @method MyClass myMethodWithReturn(MyClass $name, MyClass $slug)
 * @method static MyClass myStaticMethodWithReturn(MyClass $name, MyClass $slug)
 */
class MyMagicClass {}

### DIFF
--- a/src/Reflection/ReflectionMethodMagic.php
+++ b/src/Reflection/ReflectionMethodMagic.php
@@ -42,11 +42,11 @@ class ReflectionMethodMagic extends ReflectionMethod
 	 */
 	private $declaringClass;
 
-
 	/**
 	 * @var bool
 	 */
 	private $static;
+
 
 	public function __construct(array $settings)
 	{
@@ -57,7 +57,7 @@ class ReflectionMethodMagic extends ReflectionMethod
 		$this->returnsReference = $settings['returnsReference'];
 		$this->declaringClass = $settings['declaringClass'];
 		$this->annotations = $settings['annotations'];
-		$this->static = isset($settings['static'])?$settings['static']:false;
+		$this->static = isset($settings['static']) ? $settings['static'] : FALSE;
 
 		$this->reflectionType = get_class($this);
 	}
--- a/src/Reflection/ReflectionMethodMagic.php
+++ b/src/Reflection/ReflectionMethodMagic.php
@@ -43,6 +43,11 @@ class ReflectionMethodMagic extends ReflectionMethod
 	private $declaringClass;
 
 
+	/**
+	 * @var bool
+	 */
+	private $static;
+
 	public function __construct(array $settings)
 	{
 		$this->name = $settings['name'];
@@ -52,6 +57,7 @@ class ReflectionMethodMagic extends ReflectionMethod
 		$this->returnsReference = $settings['returnsReference'];
 		$this->declaringClass = $settings['declaringClass'];
 		$this->annotations = $settings['annotations'];
+		$this->static = isset($settings['static'])?$settings['static']:false;
 
 		$this->reflectionType = get_class($this);
 	}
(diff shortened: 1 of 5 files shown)


lookyman:
Is that static thing really a thing, and isn't there already `@static` annotation for that? Granted, I've never seen it with magic methods, but I can easily imagine something like `@method myMethod() @static` (or even `@method-static myMethod()).`

ct-jensschulze:
As you can see here at [PHPDocumentor](http://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.static.pkg.html) the @static syntax is only used inside a class. And for example phpStorm can only handle the @method static ReturnType methodName correctly for code completion.

ct-jensschulze:
And also a little example for the usage. I have a static method for instantiating the class which accepts in different classes different arguments. When i want to have the code completion support with correct argument and return types i need to add it as magic method. Works fine in phpStorm but ApiGen doesn't show this magic method.

tests/Reflections/ReflectionMethodSource/ReflectionMethod.php
```diff
@@ -15,6 +15,8 @@
  * @property-read int $skillCounter
  * @method string getName() This is some short description.
  * @method string doAnOperation(\stdClass $data, $type) This also some description.
+ * @method static string doAnStaticOperation(\stdClass $data, $type) This also some description.
+ * @method static doAnVoidStaticOperation(\stdClass $data, $type) This also some description.
```
miclf: It should be ‘doAStaticOperation’ and ‘doAVoidStaticOperation’, using ‘a’ instead of ‘an’ (generally speaking, ‘an’ is used when the following word starts with a vowel).

TomasVotruba: Indeed, cc @ct-jensschulze

src/Reflection/ReflectionMethodMagic.php
```diff
@@ -52,6 +57,7 @@ public function __construct(array $settings)
 		$this->returnsReference = $settings['returnsReference'];
 		$this->declaringClass = $settings['declaringClass'];
 		$this->annotations = $settings['annotations'];
+		$this->static = isset($settings['static']) ? $settings['static'] : FALSE;
```
TomasVotruba: Is there a case where 'static' is not present?

TomasVotruba:
Looks fine otherwise. I think it's ready to merge.

src/Reflection/ReflectionMethodMagic.php
ct-jensschulze: As i don't know exactly who is calling the ReflectionMethodMagic, just wanted to be sure it will not break existing code

TomasVotruba: Try it without check as options above and run tests. If everything works, make it the same.

But I agree, I should make it and object, so it can be sure all options are available.

ct-jensschulze: Without the check ApiGen\Tests\Reflection\TokenReflection\ReflectionFactoryTest::testCreateMethodMagic
is failing. But it seems that only this test is failing, when adding 'static' as parameter to the array it's working

TomasVotruba: I see. It's correct then, thanks for report.

TomasVotruba:
@ct-jensschulze I just merged this manually: https://github.com/ApiGen/ApiGen/commit/4435b36388997de6c18cf41ac0144d38c1aa8543

Thanks a lot for your patience. I'm sorry for keeping you waiting.

PR #535, opened 2015-01-29
Update CONTRIBUTING.md

marcovtwout:


TomasVotruba:
Great! Thank you.

Just detail, could you turn --debug into `--debug`, so it's clear it's code?

TomasVotruba:
Could you repush/squash to 1 commit?

marcovtwout:
I made this change through the GitHub interface instead of clone/fork, so this would take relatively a lot of work for such a simple change. Is the benefit worth the trouble?

TomasVotruba:
I'd like to keep simple history for others, but if it's too complicated for you, I'm fine with it.

Thanks for your work'

Issue #523, opened 2015-01-17
Display message about empty section 'todo' and 'deprecated'


grogy:
My idea:

If is section (todo or deprecated) empty then I want show a info message. What do you think? I can send pull-request (during next week)

TomasVotruba:
Hi, I'm not sure I understand. Could you describe it in more detailed way?

grogy:
I am sorry for my bad English.

**Use case:**
1. Documentation is generated with option `--todo`
2. In project no exists some todo comment.
3. I go to generated index.html.
4. I click to "todo" in header.
5. I look to empty page.
6. Here I want display information message about empty todo list.

The same use case for deprecated list.

TomasVotruba:
I see, thanks for explaining. This should be an easy one.

In such case, you'd need to modify https://github.com/apigen/ApiGen/blob/master/src/templates/default/todo.latte and https://github.com/apigen/ApiGen/blob/master/src/templates/default/deprecated.latte (as well in bootstrap theme) to check if there are any elements present.

You can prepare elements count here: https://github.com/apigen/ApiGen/blob/8202c2e46b4a0055aa6bc599713243e32e8ca2bd/src/ApiGen/Generator/TemplateGenerators/TodoGenerator.php#L76

Would you send a PR?

grogy:
Thank you. I agree, it is easy. I will send PR (during that or next week).

*(Apigen is great tool and I want contribute)*

TomasVotruba:
@grogy Great, thank you. If you have any problems or get lost, just let me know.

TomasVotruba:
@grogy Just FYI, I've decoupled default themes from master branch today. All is the same, just moved to standalone repositories_

- https://github.com/apigen/theme-default
- https://github.com/apigen/theme-bootstrap

Make PR with changes to both of them.

TomasVotruba:
Solved via https://github.com/ApiGen/ApiGen/commit/7f1123b6d971cbafc900c2b651d325cbeb0a9d8b

grogy:
Thanks :+1:

Issue #512, opened 2015-01-08
Feature for an attribute filter in the generated docs


feamsr00:
Hi,
Add some type of tag filter to support showing all (perhaps constrained by class) elements that have an, for example, @api tag.

This could be considered related to #339

-Thanks

TomasVotruba:
Hi, this could be easy to add because similar function already has @todo and @deprecated filter.

See: https://github.com/apigen/apigen/blob/master/src/ApiGen/Generator/TemplateGenerators/DeprecatedGenerator.php

It could be handy to abstract it's function to "FilteredByAnnotationGenerator", where "deprecated" would be passed.

Could you propose PR to discuss over?

TomasVotruba:
Implemented via #539

Issue #339, opened 2014-09-12
@api filter whish


malkusch:
I annotate public API classes of my library with the [`@api`](http://phpdoc.org/docs/latest/references/phpdoc/tags/api.html) annotation. Therefore I build only documentation for those classes. Before `apigen.neon` I did this with a simple Makefile target:

    ./vendor/bin/apigen.php -d . -s $$(grep -Rl api classes/ | tr '\n' ',' | head -c -1)

Before I start writing a target which generates an `apigen.neon` I was wondering if I'm the only one who uses the `@api` annotation that excessive. Maybe its worth implementing a filter option for `apigen.neon` which includes only source files with that annotation. You might consider this request as the counterpart to `internal:`. E.g:

    # include only elements marked as @api
    api: true # default [false]

TomasVotruba:
How do you like this? https://github.com/ApiGen/ApiGen/pull/539

TomasVotruba:
Feel free to comment. If relevant, I can reopen.

PR #539, opened 2015-01-29
Annotation groups generator

TomasVotruba:
Former `--todo` and `--deprecated` options now also available via `--annotation-groups` option.

```sh
--annotation-groups=todo // same as --todo
--annotation-groups=deprecated // same as --deprecated
--annotation-groups=todo,deprecated // same as --todo --deprecated
```

Now you can use any annotation you like with same output as for `--todo` and `--deprecated` options:

```sh
--annotation-groups=api,event
```

- ref #512 
- ref #339
- ref #523


tmysik:
@TomasVotruba 

> Former --todo and --deprecated options now also available via --annotation-groups option.

Please, keep it so, at least for some (longer ;) time to avoid BC break so NetBeans works correctly even with newer versions of ApiGen.

Thanks a lot!

TomasVotruba:
@tmysik I agree. They'll be removed in next major version, as described in [release process](https://github.com/ApiGen/ApiGen/wiki/Release-Process).

PR #560, opened 2015-03-02
Replace TokenReflection with own ElementReflection 

TomasVotruba:


src/Templating/Template.php
```diff
@@ -47,4 +74,34 @@ public function setSavePath($savePath)
...
+
+	/**
+	 * @return string
+	 */
+	public function render()
+	{
+		ob_start();
+		$this->latte->render($this->file, $this->parameters);
+		ob_end_clean();
+	}
```
tomaswindsor: Missing ob_get_contents() and return?

TomasVotruba: Test could tell us for sure.

TomasVotruba: @tomaswindsor Test added and fixed, thanks for notice: https://github.com/ApiGen/ApiGen/commit/040a865f0b3e554b13420cb6d62010424c7a166e

Issue #68, opened 2011-10-18
Doc Generation breaks with Umlauts/ISO8859-1-files


eth8505:
It seems that generation of docs breaks if I have umlauts (äöü) in the files (at least for me). We're (sadly) still using ISO8859-1 encoded files and have some german legacy doc blocks. 

It seems that in ApiGen\Template.php:85 the Texy instance is simply created and then forgotten about, without ever worrying about the encoding of the data, hence when $this->texy-process() calls TexyUtf::toUtf($text, $this->encoding) in texy.php:501 it still uses its internal default source encoding which is utf-8. That makes the iconv() call break since it expects UTF-8 input and gets ISO-8859-1 input.

For now, it does actually help me to get rid of all the bad doc blocks from ages ago, but I thought I'd note it here anyway :)

TomasVotruba:
@eth8505 Just curious, did this help? Are you still using non-UTF-8 encoding?

eth8505:
If I remember correctly id did help. I can't give you any up-to-date info though, since I actually left the company we used ApiGen at more than 2 years ago, sorry...

TomasVotruba:
Thanks for reply. I'm thinking about dropping this legacy support, since UTF-8 is standard nowadays, and was probably even 4 years ago.

eth8505:
I really can't say, you'd have to decide that by yourself. Where I work now, we're not actually generating API docs ourselves...

PR #562, opened 2015-03-05
drop Charset option [BC break, ref #68]

TomasVotruba:
Dropped legacy support, UTF-8 is standard nowadays.

Also should be responsibility of parser.


TomasVotruba:
Ready in 4.2

PR #565, opened 2015-03-08
Exclude/source over skipDocPath

TomasVotruba:
`skipDocPath` option allows generation of doc, where only method names are displayed with no more details.

In case of skipping generation of whole lib (e.g. nette\nette package only for Nette\Object class), it's recommended including Nette\Object.php in sources. In this way, all methods with full descriptions will be allowed.

For excluding of multiple dirs/files `--exclude` option should be used.


TomasVotruba:
Wrong target branch, closing.

PR #582, opened 2015-04-06
use Nette 2.3

TomasVotruba:


GrahamCampbell:
Using a version constraint of `~2.2` means you were already using 2.3 though?

TomasVotruba:
Yep, but 2.2 was allowed. Here is 2.3 features exclusively.

GrahamCampbell:
Right. :)

PR #585, opened 2015-04-06
Deprecated generators

TomasVotruba:
Didn't bring much value to generated docs


fprochazka:
Why? I liked the sitemap :-/

TomasVotruba:
It's not so relevant to the docs. Also still to big to maintain for me.

I moving toward plugins like composer has, to allow add custom generators, if you really need it.

fprochazka:
Well, I just fail see the benefit of removing something that is working...

TomasVotruba:
Also it's not suitable for these kinds of projects, quoting [Google Docs](https://support.google.com/webmasters/answer/156184?hl=en).

> Well, I just fail see the benefit of removing something that is working...

Development process is not in "working" state yet. I'd really appreciate bugfixes to make closer to that state.

PR #587, opened 2015-04-07
Update composer

emulienfou:


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-	"name": "apigen/apigen",
+	"name": "emulienfou/apigen-markdown",
```
GrahamCampbell: ?

composer.json
```diff
@@ -30,6 +30,7 @@
 
 		"apigen/theme-default": "~1.0.1",
 		"apigen/theme-bootstrap": "~1.1.2",
+		"emulienfou/theme-markdown": "dev-master",
```
GrahamCampbell: ?

GrahamCampbell: Also, dev-master should NEVER be used as a version constraint.

PR #596, opened 2015-04-13
update on use method's

Tjoosten:
some use elements declared but never used. So i remove them.


TomasVotruba:
@Tjoosten Thank you

PR #599, opened 2015-04-17
Composer: fix version constraints

DavidPrevot:
Use caret instead of tilde to allow any version of Symfony component up
to 3.0 (instead of limiting symfony/options-resolver to 2.7).

https://getcomposer.org/doc/01-basic-usage.md#next-significant-release-tilde-and-caret-operators-


TomasVotruba:
Thanks for suggestion. I've implemented it in the master.

PR #608, opened 2015-05-24
Contracts

TomasVotruba:


tests/Templating/Filters/AnnotationFiltersTest.php
```diff
@@ -55,13 +55,13 @@ public function testAnnotationFilterWithCustom()
 	}
 
 
+	/**
+	 * @deprecated since 4.2. To be removed in 5.0.
```
GrahamCampbell: The tests aren't part of the public api though?

GrahamCampbell: They're in the autoload dev section.

TomasVotruba: I don't understand, could you elaborate?

GrahamCampbell: I'm saying what's the point in deprecating this. It's not a breaking change to remove this given that no external library is allowed to be using this class.

TomasVotruba: I see. I think it can be removed then.

PR #632, opened 2015-07-31
FileSystem

Modicrumb:
Changed one / to the Directory Separator PHP constant allowing for this method to potentially work on both windows and unix filesystems


Modicrumb:
This actually fails a test on a windows machine, and I feel the author already took windows into account when creating this, after further research this pull is not necessary.

Issue #520, opened 2015-01-14
Tempfiles not removed on exit


MyIgel:
After running apigen, temp files are created at /tmp/_apigen.temp/ but not removed after execution.
This makes it impossible for other users on multi user systems to run phpgen because they can't be deleted

```
PHP Warning:  unlink(/tmp/_apigen.temp/cache/latte/phar-src-templates-default-elementlist-latte-d3f4e2a491a20b9dce9feb7f60a2f85c.php): Permission denied in phar:///~/bin/apigen/src/ApiGen/FileSystem/FileSystem.php on line 64

Warning: unlink(/tmp/_apigen.temp/cache/latte/phar-src-templates-default-elementlist-latte-d3f4e2a491a20b9dce9feb7f60a2f85c.php): Permission denied in phar://~/bin/apigen/src/ApiGen/FileSystem/FileSystem.php on line 64
PHP Warning:  unlink(/tmp/_apigen.temp/cache/latte/apigen-phar-src-templates-default-layout-latte-d7032cd44fe6ff49bbb4964b02cf15cf.php): Permission denied in phar://~/bin/apigen/src/ApiGen/FileSystem/FileSystem.php on line 64

Warning: unlink(/tmp/_apigen.temp/cache/latte/apigen-phar-src-templates-default-layout-latte-d7032cd44fe6ff49bbb4964b02cf15cf.php): Permission denied in phar://~/bin/apigen/src/ApiGen/FileSystem/FileSystem.php on line 64
PHP Warning:  unlink(/tmp/_apigen.temp/cache/latte/apigen-phar-src-templates-default-overview-latte-50bc665319409d0c611e3d7240df1f17.php): Permission denied in phar://~/bin/apigen/src/ApiGen/FileSystem/FileSystem.php on line 64

[...]
```

A fast workaround is to use TMPDIR=~/tmp apigen [...] but it should be fixed

TomasVotruba:
I see. Do you have suggestions?

MyIgel:
Maybe append a random suffix to the temp folder and make shzre to delete it when apigen ends?

TomasVotruba:
I've added temp files cleanup [after run](https://github.com/ApiGen/ApiGen/commit/f1c36c004a06c9e4a951b04b7385bd4f068a910c#diff-2910e1538b17d63253104b38e9558defR40).

Could you please validate this solution by using `composer require apigen/apigen:@dev`?

TomasVotruba:
[...]

PR #650, opened 2015-10-06
real fix for issue #520

POPSuL:


jadb:
Thanks @POPSuL! Could you target 4.2 as well?

POPSuL:
@jadb yes, I can :)

PR #654, opened 2015-10-10
Update ./generate-api.sh

deizel:
I've cleaned up the `./generate-api.sh` script:

- Less Git voodoo (just clone/commit/push, rather than creating a new repo every time)
- No longer force pushes branch and wipes out history
- More portable (others can just run existing script with different environment variables)
- Doesn't change your global git config when testing locally

You'll notice I also excluded `.git` from the purge directories command:

> Destination is not empty. Do you want to erase it? [yes]

I don't see why anyone would want that so hopefully nobody relies on this.

This should likely be called from an `after_deploy` in the Travis config.


jadb:
Sweet! Let's finalize/test this tonight and get it released :)

deizel:
Going to retarget this against `4.1` actually.

Issue #652, opened 2015-10-08
List of known projects ApiGen fails to build


deizel:
Ran ApiGen against 124 repositories ([script](https://gist.github.com/deizel/2795a26bf18ad30dcba9)) and this is the list of packages that failed to build.

4.1.x is looking in good shape while 4.2.x needs some fixes before it can be considered stable.

## v4.1.0 (phar) & v4.1.1 (git)

Refs [#631](https://github.com/ApiGen/ApiGen/issues/631): (fix currently in `4.1` branch, to be released in `v4.1.2` tag, refs #653 - thanks @jadb)

```
Argument 2 passed to ApiGen\Generator\Resolvers\ElementResolver::resolveContextForClassProperty()
must be an instance of ApiGen\Reflection\ReflectionClass, instance of ApiGen\Reflection\ReflectionFunction given
```

- [x] ~~beberlei/assert~~

Refs #629: (see old issue #272 - fork is no longer available)

```
'Trait method "X" was already imported.' in vendor/andrewsville/php-token-reflection/TokenReflection/ReflectionClass.php:734
```

Fixed in (aptly numbered) #666 - phpunit and folowing projects passing: (@deizel)

- [x] ~~gsouf/UForm~~ (manually verified, trait methods are displaying now)
- [x] ~~league/climate~~
- [x] ~~pomm-project/Foundation~~
- [x] ~~react/promise~~
- [x] ~~symfony/symfony~~ (?) - ~~takes too long to generate, will confirm later~~ (took 11 minutes but it works)

## v4.2.0-dev (dev-master)

No tickets matching error:

```
Fatal error: Call to a member function getDefaultValueDefinition() on null
in src/Parser/Reflection/ReflectionParameter.php on line 62
```

Fixed in #665 (thanks @Phillaf):

- [x] ~~beberlei/assert~~
- [x] ~~cakephp/cakephp~~
- [x] ~~codeception/codeception~~
- [x] ~~codegyre/robo~~
- [x] ~~fzaninotto/faker~~
- [x] ~~guzzlehttp/guzzle~~
- [x] ~~league/climate~~
- [x] ~~nesbot/carbon~~
- [x] ~~nette/utils~~

No tickets matching error: (fix in #663 - slated for `v4.2` tag - thanks @POPSuL)

```
Fatal error: Call to a member function isDeprecated() on null
in src/Parser/Reflection/ReflectionElement.php on line 129
```

- [x] ~~nikic/php-parser~~
[...]

PR #664, opened 2015-10-11
Parser exceptions

deizel:
`ParserExceptions` were not being handled so the `GenerateCommand` would fail.

Also, messages like "Unexpected token found." are not very useful for tracking things down.

Now with `--debug` flag, you will get output such as:

```
Parse error: "There were following reasons for this exception:
Thrown when working with file "tests/_fixture/InterfaceWithSemiReservedMethodName.php" token stream.
The cause of the exception was the T_UNSET token (line 4) in following part of the source code:

 1: <?php
 2: interface InterfaceWithSemiReservedMethodName
 3: {
*4:     public function unset();
 5: }
 6: "
```

Also, the API will continue generating until 100%, rather than halting abruptly.

Refs #652

### DIFF
--- a/src/Console/Command/GenerateCommand.php
+++ b/src/Console/Command/GenerateCommand.php
@@ -268,12 +268,18 @@ class GenerateCommand extends AbstractCommand
     {
         /** @var FileProcessingException[] $errors */
         foreach ($errors as $error) {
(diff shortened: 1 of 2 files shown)


jadb:
:+1:

PR #667, opened 2015-10-11
Trait methods

deizel:
Follows #666


jadb:
Maybe break this big method now in smaller ones?

PR #676, opened 2015-11-01
make temp dir user dependent

remicollet:
When installed in a shared localtion, it can be use by various users.
Each user need a different cache directory to avoid permission denied error on cleanup.


jadb:
:+1:

PR #693, opened 2015-12-14
'branch' is ignored when 'tags' are enabled

rindeal:
https://docs.travis-ci.com/user/deployment#Conditional-Releases-with-on%3a

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,6 @@ deploy:
   file: apigen.phar
   on:
     repo: ApiGen/ApiGen
-    branch: 'master'
     tags: true
 
 after_deploy:


jadb:
Thanks!

jadb:
I hope I wasn't too click-happy. This will trigger for tags and `master`, correct?

rindeal:
Incorrect, it will trigger whenever you tag some commit, no matter which branch it's currently in. If you want to limit it just to the `master` branch, then tag only commits in that branch ;-). I don't think there is currently any native Travis way of preventing it. If you were really serious about it though, you could utilize curl and GitHub API.

Anyway, my commit hasn't changed any functionality whatsoever.

PR #706, opened 2016-02-24
composer: unpined Latte, bug was removed in 2.3.6

dg:
related to d1e3d98de


jadb:
Thanks @dg!

@deizel can you run it on the full test suite you have?

deizel:
Looks good, tests passing after merging to master alongside #703

PR #705, opened 2016-02-24
fixed relative path resolver on Windows

dg:

PR #707, opened 2016-02-25
fixed TreeGenerator, now really generates trees

dg:
Currently it generates flat structure.


jadb:
Tests are failing here :cry:

dg:
I'll check it

dg:
@jadb tests fixed.

Also fixed test fail introduced in https://github.com/ApiGen/ApiGen/pull/705

dg:
Btw, there is already cca 10 tests that fails on Windows because of different directory separator

dg:
Hmmm, there is still one failing test in some environment `ApiGen\Tests\Generator\TemplateGenerators\TreeGeneratorTest::testCanBeProcessed
Failed asserting that true is false.` but I have absolutely no idea what is does and how to fix it. I am throwing in the towel.

PR #709, opened 2016-02-25
isUri replaced with isUrl because only URLs are clickable

dg:
Otherwise ` {@link Presenter::$invalidLinkMode}` will generate `<a href="Presenter::$invalidLinkMode">Presenter::$invalidLinkMode</a>`


jadb:
Thanks!

PR #723, opened 2016-07-10
Fix TreeGenerator not making a tree.

### LINKED ISSUE #569: Tree does not display as expected
Using 4.1-beta I have an issue generating tree structure, the tree isn't well formed i.e missing branches compared to tree that I see inside class.

## Complete tree structure:
tree_apigen

## Partial tree visible from class:
tree_inside_class_apigen

In tree structure Model branch is not expanded as in class tree.

markstory:
Previously the TreeGenerator created a flat list. This made for a very boring 'tree' as there was only one level to it. These changes build on those in #707, and updates the test to cover the defect.

Fixes #569

### DIFF
--- a/src/Generator/Markups/MarkdownMarkup.php
+++ b/src/Generator/Markups/MarkdownMarkup.php
@@ -51,7 +51,7 @@ class MarkdownMarkup implements Markup
      */
     public function block($text)
     {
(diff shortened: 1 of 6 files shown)


markstory:
The failing tests failed both before and after my changes.

jadb:
Thanks @markstory and @antograssiot - sorry for the delay! Trying to catch up on things.

markstory:
:+1:

Issue #672, opened 2015-10-19
"Invalid value definition" when using ::class for constants


kdambekalns:
Code like `const foo = \Acme\Com::class;` cannot be documented.

Similar to #621 and #627.

TomasVotruba:
Resolved via #726

Issue #627, opened 2015-07-17
Generate fails when using class constants with array value


etienneq:
This issue applies to Apigen 4.1.1.
One of my classes has a constant like this: `const FOO = [ 1 => 123, 2 => 456];`

When calling apigen generate the following error message is given: `Value definition is empty.`

It's the same with: `const FOO = array(1 => 123, 2 => 456);`

Only `const FOO = 'bar';` works.

I suppose that's because Apigen doesn't yet support PHP 5.6 language features?

jadb:
That's correct. Do you want to try adding support for that? Or maybe a failing test to get things started?

jadb:
Similar to #621 - but this one with arrays.

TomasVotruba:
Resolved via #726

Issue #621, opened 2015-06-28
Invalid value definition when using constant expressions


BenMorel:
PHP 5.6 supports [constant expressions](http://php.net/manual/en/migration56.new-features.php), but ApiGen fails with _Invalid Value Definition_ when using string concatenation in constants:

```
const VALUE = 'foo' . 'bar';
```

TomasVotruba:
Resolved via #726

StormBytePP:
This one is still failing for me in code such as:
class A {
      const TMPDIR = __DIR__."/../../../tmp";
}

Can you reopen?

TomasVotruba:
Sure, I will make sure to look on that after reflection switch.

Thanks for providing failing code snippet :+1:

TomasVotruba:
Probably missing support at [BetterReflection](https://github.com/roave/BetterReflection)

- `CompileNodeToValue::__invoke()`

Add in the end sth like:

```php
if ($node instanceof Node\Scalar\MagicConst\Dir) {
    return $this->compileDirConstant($node, $context);
}
```

TomasVotruba:
Fix in progress: https://github.com/Roave/BetterReflection/pull/280

TomasVotruba:
And done!

PR #726, opened 2016-07-12
Changed php-token-reflection dependency

POPSuL:
Hello!

I created a fork of php-token-reflection (https://github.com/POPSuL/PHP-Token-Reflection) which I'm going to support as I can.

At this moment I collected all changes from that repository https://github.com/kornrunner/PHP-Token-Reflection (many fixes for PHP 5.6) and I added basic support for constants expressions (like const A = 4 / 2) and array constants (like const B = [1, 2, 3]).

Before:

ApiGen returns error like "Value definition is empty" for code like this:

```php
class A {

    const B = [1, 2, 3];
}
```

After:
Everything works perfectly! :+1:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
 	],
 	"require": {
 		"php": ">=5.5",
-		"andrewsville/php-token-reflection": "^1.4",
+		"popsul/php-token-reflection": "^1.5",
 		"apigen/theme-bootstrap": "1.1.*",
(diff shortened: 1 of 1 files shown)


POPSuL:
This PR also affects #672, #621 and #627

jadb:
@POPSuL thanks!

TomasVotruba:
Thank you @POPSuL

Issue #724, opened 2016-07-10
apigen generate throwing ErrorException: Nette\DI\ServiceDefinition::getImplementType() is deprecated


ApeWare:
I get an exception when I run `aigen generate`. The exception appears to be coming from nette/application.
## Debug Output

I get same error when I run `vendor/bin/apigen generate --debug`

``` BASH
╭─vagrant at homestead in ~/Projects/v3 on an-feature ⚡
╰─± apigen generate --debug
ErrorException: Nette\DI\ServiceDefinition::getImplementType() is deprecated, use getImplementMode() in /home/vagrant/Projects/v3/vendor/nette/di/src/DI/ServiceDefinition.php:321
Stack trace:
#0 [internal function]: Tracy\Debugger::errorHandler(16384, 'Nette\\DI\\Servic...', '/home/vagrant/P...', 321, Array)
#1 /home/vagrant/Projects/v3/vendor/nette/di/src/DI/ServiceDefinition.php(321): trigger_error('Nette\\DI\\Servic...', 16384)
#2 /home/vagrant/Projects/v3/vendor/kdyby/events/src/Kdyby/Events/DI/EventsExtension.php(301): Nette\DI\ServiceDefinition->getImplementType()
#3 /home/vagrant/Projects/v3/vendor/kdyby/events/src/Kdyby/Events/DI/EventsExtension.php(131): Kdyby\Events\DI\EventsExtension->autowireEvents(Object(Nette\DI\ContainerBuilder))
#4 /home/vagrant/Projects/v3/vendor/nette/di/src/DI/Compiler.php(231): Kdyby\Events\DI\EventsExtension->beforeCompile()
#5 /home/vagrant/Projects/v3/vendor/nette/di/src/DI/Compiler.php(160): Nette\DI\Compiler->generateCode()
#6 /home/vagrant/Projects/v3/vendor/nette/bootstrap/src/Bootstrap/Configurator.php(275): Nette\DI\Compiler->compile()
#7 [internal function]: Nette\Configurator->generateContainer(Object(Nette\DI\Compiler))
#8 /home/vagrant/Projects/v3/vendor/nette/di/src/DI/ContainerLoader.php(118): call_user_func_array(Array, Array)
#9 /home/vagrant/Projects/v3/vendor/nette/di/src/DI/ContainerLoader.php(82): Nette\DI\ContainerLoader->generate('Container_b3811...', Array)
[...]

PR #728, opened 2016-07-22
Remove incompatible dependency

malkusch:
ApiGen currently doesn't play well with nette-2.4:

> Nette\DI\ServiceDefinition::getImplementType() is deprecated

This prevents generating documentation or running the test suite.

This commit would remove the offending dependency. Tests are
failing, but I couldn't see if this is a regressen, as before
the tests were failing because of said incompatibility.

With this commit the tests are at least running again and the
documentation can be generated as well.

See also: ApiGen/ApiGen#724 Kdyby/Events#99

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,9 @@
 		"squizlabs/php_codesniffer": "~2.0",
 		"phpunit/phpunit": "~4.0"
 	},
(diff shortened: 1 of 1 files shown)


jadb:
Thanks!

PR #735, opened 2016-08-25
Update box.json according to changes made in #726

ramsey:
This updates the box.json configuration (according to changes made in #726) so that Phar files may be created again.


jadb:
Thanks @ramsey!

Issue #679, opened 2015-11-06
Overwrite destination without interaction


marcovtwout:
Is there a setting so destination directory is always emptied first without prompting?

mjh75:
+1. Having to manually delete the files is frustrating. From what I have read there used to be a way until someone shot himself in the foot and then it was removed. Config file or command line would be fine (both would be best).

marcovtwout:
For extra safety, a check could be performed for the existance of certain generated files before the directory is emptied.

mjh75:
Or if the config file has overwrite: Yes then assume the author wrote it on purpose. If I shoot myself in the foot then it is my problem. :)

ApeWare:
+1 I can delete the directory first with gulp then run the apiGen command but it seems practical that this should be able to be passed as a parameter.

jsifuentes:
+1, currently dealing with this now.

jadb:
Anyone wishes to take a shot at it?

TomasVotruba:
Resolved by #738

PR #738, opened 2016-08-26
Added support for --overwrite (-o) option

POPSuL:
Solve for issue #679


jadb:
No tests? 😝

POPSuL:
@jadb this case can not be tested with phpunit :confused: 
I tried, but I dont know how to do it. It works, I swear! :joy:

jankal:
@TomasVotruba do we need tests for that

PR #741, opened 2016-08-29
Fixed #740 - functions can not be a type definition

### LINKED ISSUE #740: Invalid return typehint resolving
In some cases apigen incorrectly resolves return typehints.

Example:

```php
function int()
{
    //
}

class A
{

    /**
     * @return int|int[]
     */
    public function foo()
    {
        return 0;
    }
}
```

Actual result:
image

But, function is cannot be a return value. And expected result should be `integer|integer[]`.

POPSuL:
+ added support for array-notation for simple types

### DIFF
--- a/src/Templating/Filters/Filters.php
+++ b/src/Templating/Filters/Filters.php
@@ -58,6 +58,12 @@ abstract class Filters
         ];
 
         // Simple type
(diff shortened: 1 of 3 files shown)


jadb:
Sweet! Do you have time for tests as well?

POPSuL:
@jadb i will add tests later :)

PR #769, opened 2017-01-12
Add support for v3 version of Symfony Components

ek9:
Add support for v3 version of Symfony Components. Update the test so it can be run with v2 and v3 version of `symfony/console`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,10 +23,10 @@
 		"nette/di": "^2.3",
 		"nette/neon": "^2.3",
 		"nette/reflection": "^2.3",
-		"symfony/console": "^2.7",
-		"symfony/event-dispatcher": "^2.7",
-		"symfony/options-resolver": "^2.7",
-		"symfony/yaml": "^2.7",
+		"symfony/console": "^2.7|~3.0",
+		"symfony/event-dispatcher": "^2.7|~3.0",
+		"symfony/options-resolver": "^2.7|~3.0",
+		"symfony/yaml": "^2.7|~3.0",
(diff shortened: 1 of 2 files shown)


ek9:
This fixes #770 

Please let me know if there is something I might have missed, but all the tests run fine in both versions of `symfony/console`. Otherwise, I did not have to update anything else. Unless there are things not convered by tests, this should be safe to merge.

ek9:
ping @jadb @POPSuL , could we merge this in for Symfony3 support?

ek9:
Any comments or status on the PR? :)

jankal:
Close and reopen for CI run again.

jankal:
My be closed in favor of #789

ek9:
Just a quick ping to @jankal, my fix had support for both symofny 2.8 and 3.0 series. The older 4.1.* version have problems which require you to use master for symfony 2.8 LTS. I believe the 4.2 series should support both (2.8 and 3.0), especially that now PHAR support is dropped and there is no way to use `apigen` without it dictating the versions of components to use. I would suggest to go symfony 3.0 only on 4.3 in the future.

jankal:
@TomasVotruba ping

ek9:
Perhaps it would be more appropriate to discuss this on https://github.com/ApiGen/ApiGen/issues/770 issue. I've made separate comment there + explained the use case.

TomasVotruba:
There have been many BC breaks, to revive the project.
I'm here to focus on 5.0 release. I don't have time to handle anything more.

PR #781, opened 2017-03-11
Upgrade phpunit to ~6.0

jankal:
Following #782.


jankal:
closed in favor of #788

Issue #782, opened 2017-03-11
Make future-proof


jankal:
# Future-proof is key for not burying this project any time soon

So hello. As I've been using ApiGen privately for 1.5 years now, I really see the need to modernize it. Else it will be dead (as @phenixiim already realized in #771). Therefor we probably need to drop backwards compatibility for PHP 5 for the next releases.

For all the BC lovers: Users should still be able to use an older version supporting PHP 5.

I will from now on start with contributing to this project (yep: @ek9 #779). For this the following list of tasks need to be done to get this future-proof.

Maybe someone part of the org (@jadb @kukulich) could add a [GitHub project](https://github.com/ApiGen/ApiGen/projects) for that and import this list then.

Moved list to repo (https://github.com/ApiGen/todolist/blob/master/new-version.md)

For any other ideas: feel free adding them as a comment so the community and also the contributors (including me) can work on it.

I really don't like the idea of burying this project as it has a very huge codebase and is the outcome of many peoples work with as I see probably thousands of lines of code. Haveing roud 580 tests is very impressive. We will keep this up and make it better.

Pepole who volunteered for contributing:
 - @TomasVotruba 
- @nozavroni #495

TomasVotruba:
#### Thanks for putting this down!

I'd also recommend to bumping to PHP 7.1. 
Really helped me with not burying my projects.

Can help if needed. I'm quite fast with it :)

jankal:
@TomasVotruba Do you have any permissions on this repo?

TomasVotruba:
@jankal I don't this so.

jankal:
@TomasVotruba see #779

TomasVotruba:
Just a thought. I'm thinking about removing Markdown, since it brings so various syntax problems and **encourages people to actually use Markdown in PHP comments**.

I consider this mixing of concerns. Docblock should be used for information, not for writing posts - blogs are for that. It would lighten this project a bit more.

What do you think? Do you have some use cases where this is needed?

TomasVotruba:
Another idea is for dropping **tree** support. I never really used and it just show all classes. I think it's same use as showing all posts on blog. I look for particular class and it's relations, not to see it all.

What do you think?

jankal:
@TomasVotruba Will put this into a GitHub Proejct now.

ek9:
Small comment regarding version bumps:

[...]

PR #783, opened 2017-03-12
PHP 7 required!

jankal:
Following #782 I updated README and `.travis.yml` for the new dependency on PHP 7.


jankal:
Why is Travis using PHPUnit 6.0.8 even it shows that PHPUnit 4.8.35 was installed?

jankal:
May be closed in favor of #788

PR #787, opened 2017-03-12
Bump to PHP 7.1

TomasVotruba:

Issue #780, opened 2017-03-06
Downloaded v4.1.2 but ApiGen -V says 4.1.0


Yousha:
I just downloaded **ApiGen v4.1.2** from here https://github.com/ApiGen/ApiGen/releases/download/v4.1.2/apigen.phar but when I use **php apigen.phar -V** command it says **ApiGen version 4.1.0**.

Also issue #732 is not fixed in this version(v4.1.2)
`php apigen.phar selfupdate`
> The JSON data did not pass validation.

jankal:
Closed. PHAR not supported since #788.

PR #775, opened 2017-02-05
herrera-io/phar-update is no longer supported

zonuexe:
https://packagist.org/packages/herrera-io/phar-update
> This package is abandoned and no longer maintained. No replacement package was suggested.

[deployer/phar-update](https://github.com/deployphp/phar-update) seems to be actively supported.

Issue #773, opened 2017-01-17
herrera-io/phar-update dependency abandoned


ek9:
`herrera-io/phar-update` dependency abandoned, need to replace with something maintained

jankal:
Closed. PHAR not supported since #788.

PR #788, opened 2017-03-12
Drop phar

TomasVotruba:
## Why?

This requires too much work and causes too many issues - see those 10 bellow.
Rather than playing with this, we should focus on reflection and PHP 7+ features.

Depends on: #787 

Resolves: 

- #780
- #773
- #761
- #760
- #733
- #732
- #675
- #638
- #775

### Dependencies Cleanup

image


jankal:
Nice.  @TomasVotruba Thanks for removeing PHAR support. This will make deployment a lot easier (almost no deployment...).

TomasVotruba:
I've noticed PHPUnit too late. What this is not merged?
To bad.

Yep, it should bring more peace in here :) I know I spent weeks on it here.

jankal:
Nah. I think it's good enough like this.

May I merge your version?

TomasVotruba:
Awesome!

TomasVotruba:
Could you also close those issues that I mentioned?

jankal:
Yep I will.

PR #790, opened 2017-03-13
Drop download option (mobile net is fast nowadays)

TomasVotruba:
I think we can drop this feature, since using ZIP is kinda useless today. Internet is just fine, everywhere.

PR #791, opened 2017-03-13
Drop todo (already deprecated)

TomasVotruba:
This is rather note for developer than documentation value.


fschlich:
I think todo and deprecated are useful markers - isn't apidoc primarily for developers (of depending projects, sure, but also for new contributors or people not yet very familiar with the history and code)

jankal:
Merged cause it's no merge into master yet. Discussion fro all EOL features in #790

PR #792, opened 2017-03-13
Drop tree

TomasVotruba:
This is rather nice-to-see once feature than useful one. I never really used and it just show all classes. I think **it's same use as showing all posts on blog**. I look for particular class and it's relations, not to see it all.


jankal:
Have no opinion on that myself.

Do @kukulich @POPSuL @ek9 have any opinions about this?

POPSuL:
I did not even know about this and did not use this thing.
If this is the same as the one in the left pane (separate page with tree of namespaceses with classes) - it is not necessary IMHO.
If I'm wrong - show a screenshot of what you mean. I currently can't see what this option does :(

jankal:
@POPSuL I think you meant the same thing as I also meant ;).

So this is ACCEPTED and will be MERGED.

TomasVotruba:
@POPSuL Same for me. Yep it's pretty similar to the structure in the left :) 

Thank you guys.

PR #794, opened 2017-03-13
Drop deprecated (already deprecated)

TomasVotruba:


jankal:
When we drop support here, we need to also drop support on the themes.

fschlich:
I think deprecated is a useful feature, both for people working on code that uses your library to update before it stops working, as well as for the library devs themselves to remember to get back to removing deprecated features after good while.

Please don't remove unless it's actually a burden to maintain!

TomasVotruba:
This feature is covered by annotation-group

TomasVotruba:
Thank you

Issue #772, opened 2017-01-17
Working w/php Markdown Extra


CNSKnight:
I've got blocks that might include a replacer prompt that looks like an HTML \<tag\>
```
    /**
    * pulls in any default <MODULE_NAME>.css file in the modules templates/css/
    * 
```
But I'm not getting the htmletities conversion as one might expect.
![apigen-markdown-ss](https://cloud.githubusercontent.com/assets/220025/22029585/fd4f9d0c-dca8-11e6-80da-95de342a2f71.png)

While more cluttered reading inline, I tried escaping \\<like this\\> and got this:
![apigen-markdown-ssb](https://cloud.githubusercontent.com/assets/220025/22029864/d32c1b44-dca9-11e6-98de-75d1be5a3528.png)

Any further suggestions?

jankal:
Markdown has been removed by #795. Therefor this issue is nonexistent right now.

Issue #718, opened 2016-05-06
Support for @example


jaschweder:
This is a improvement that I think is important, see bellow.

Sometimes is important to show a example of use of the function/class/etc, the @example tag will be perfect for this, using single and multi lines, see some examples:

``` php
class Bank {

    /**
    *  @param $account The new account
    *  @example
    *   $account = new Account;
    *   $account->setNumber(12356);
    *   $bank = new Bank;
    *   if($bank->createAccount($account)){
    *       // everything is ok
    *   }
    */
    public function createAccount(Account $account)
    {
        if($this->validator($account)){
            $this->accounts->add($account);
            return true;
        }
        return false;
    }
}
```

In the API generated files this can be show has any other tag but with code highlight :smile:

jadb:
Sorry for the late reply.

You can already do this using back-ticks:

```php
class Bank {

    /**
    *  
    * ```
    *   $account = new Account;
    *   $account->setNumber(12356);
    *   $bank = new Bank;
    *   if($bank->createAccount($account)){
    *       // everything is ok
    *   }
    *```
    *
    *  @param $account The new account
    */
    public function createAccount(Account $account)
    {
        if($this->validator($account)){
            $this->accounts->add($account);
            return true;
        }
        return false;
    }
}
```

Did I misunderstand?

jaschweder:
@jadb I believe this is the point, but something enhancement can be the code highlight in the code generated, don't you thing ?

I make some proposal commit's here

jadb:
You mean having the sampled code highlighted when rendered?

jaschweder:
@jadb code highlight in the example code and code format.

Holt59:
Maybe this is related to my bootstrap theme, but this does not work because the generated code is inlined, so it does not render properly:

```html
<pre>$account = new Account; $account->setNumber(12356); $bank = new Bank; if($bank->createAccount($account)){ // everything is ok }</pre>
```

Instead of:
```html
<pre>$account = new Account; 
$account->setNumber(12356); 
$bank = new Bank; 
if($bank->createAccount($account)){ 
    // everything is ok 
}</pre>
```

ek9:
The example provided by @jadb will probably no longer work as markdown support was removed. @TomasVotruba @jankal  can you confirm this? Also what are your thoughts about this issue? Is this something we plan to have for 5.0?

TomasVotruba:
@ek9 The @example annotation should work for source code highlighting even without Markdown.

[...]

Issue #671, opened 2015-10-18
Code example formatting issue.


WitteStier:
Code examples like below 

``` php
 $data = [
     'user_id'           => 1
     'user_email'        => 'admin@example.com'
     'user_password'     => '***',
     'person_id'         => 1,
     'person_first_name' => 'John',
     'person_last_name'  => 'Doe',
 ];
```

are rendered as 

``` php
$data = [ 'user_id'           => 1 'user_email'        => 'admin@example.com' 'user_password'     => '***', 'person_id'         => 1, 'person_first_name' => 'John', 'person_last_name'  => 'Doe',
];
```

olvlvl:
Have you tried _fencing_ your code? https://help.github.com/articles/github-flavored-markdown/#fenced-code-blocks

jadb:
@olvlvl thanks for replying :+1: 

@WitteStier have you tried fencing as suggested above?

ramsey:
#647 appears to be the same issue.

I'm also noticing this in ApiGen 4.1.0, and I am using fencing.

WitteStier:
Sorry for my late response. Yes i've "fenced" the code.

onet4:
Commenting out these lines fixes this issue but it may break something else as I'm not sure what the lines are for.

https://github.com/ApiGen/ApiGen/blob/4.1/src/Templating/Filters/UrlFilters.php#L203-L207
```
		$long = preg_replace_callback('~(?:<(code|pre)>.+?</\1>)|([^<]*)~s', function ($matches) {
			return ! empty($matches[2])
				? preg_replace('~\n(?:\t|[ ])+~', ' ', $matches[2])
				: $matches[0];
		}, $long);
```

jankal:
Markdown has been removed by #795. Therefor this issue is nonexistent right now.

Issue #647, opened 2015-09-30
Markdown code block formatting problems


cnanney:
The following code block
```php
$user = new User('dp');
try {
    $user->authenticate('credential', $input);
    redirect($user->get('auth.redirect_url'));
}
catch (UserLoginException $e) {
    redirect($e->getData()['redirect_url']));
}
```
Renders as
![image](https://cloud.githubusercontent.com/assets/195134/10205971/04983d70-678a-11e5-9506-ef81da446ef3.png)

The code inside the try and catch blocks gets flattened to one line. Same issue with and without syntax highlighting, so must be the markdown parser. Hopefully will be fixed by #634.

PR #795, opened 2017-03-13
Drop Markdown

TomasVotruba:
### Why?

This brings various syntax problems and encourages people to actually use Markdown in PHP comments. I see this as a bad practice and source of great confusion what is formatted and how.

I consider this mixing of concerns. **Docblock should be used for information**, not for writing posts - blogs are for that. It would lighten this project a bit more.

I recall this brought so many issue with spaces, `<p>` left overs, lines vs. block and crippled code highlighting.

### Would resolve confusions and issues

- https://github.com/ApiGen/ApiGen/issues/772
- https://github.com/ApiGen/ApiGen/issues/426
- https://github.com/ApiGen/ApiGen/issues/671
- https://github.com/ApiGen/ApiGen/issues/647
- https://github.com/ApiGen/ApiGen/issues/671
- https://github.com/ApiGen/ApiGen/issues/634
- https://github.com/ApiGen/ApiGen/issues/543
- https://github.com/ApiGen/ApiGen/issues/426
- https://github.com/ApiGen/ApiGen/issues/541
- https://github.com/ApiGen/ApiGen/issues/597

### @example as specific and clear alternative

As already requested, annotation @example might do the trick just fine with much more predictable behavior
https://github.com/ApiGen/ApiGen/issues/718


POPSuL:
Maybe the engine for rendering needs to be replaced? E.g. parsedown.
I'm against removing markdown support :(

TomasVotruba:
We are in state of dying package with over 10 issues with this section. 
What value brings Markdown (or any -down) over missing PHP 7 reflection features?

POPSuL:
Many issues caused by ugly markdown library which uses.
If we remove the markdown support - PHP7 support will not be improved, but the backwards compatibility will be broken.
I chosen ApiGen because it support markdown syntax for documenting, and I have many source code with markdown in phpDoc. I think I'm not the only one.

TomasVotruba:
What is your use case?
What about event access for you to do anything you need?

TomasVotruba:
Thing is about the maintenance overhead. The more there are feature, the less there is attention for them all.

zonuexe:
Markdown is an important function, but I feel it is more buggy. I agree with this proposal.

jankal:
Please get in touch with @kukulich if you need any repo write permission so you may fix the markdown engine. I personally also only consider this a real feature when no problems rising up in futur version will be related to it.

So when you do some work on it: please use a highly stable and supported rendering engine. As I see there's alreadfy been a little PR war about Markdown enignes. I don't want this again. Else we will drop it completly (this pull will tghen be accepted). 

Now on hold @POPSuL

POPSuL:
Also @jadb and @deizel should be invited to this dispute.
IMHO - markdown is a killer feature for apigen. Nobody wants write HTML markup in own PHP sources (its ugly). Markdown have perfect readability in source code and it should be presented in apigen.
michelf/php-markdown - is not better library for md rendering, there are alternatives.
As @TomasVotruba says "we have many troubles with apigen, we don't have support PHP7, we have many issues about rendering (md rendering)". But how does md support related to PHP7 support? 
If we remove md support and close 10-20-30 issues, implementation for PHP7 support will not become easer.
I propose replace md-renderer with better library...

TomasVotruba:
@POPSuL 

> If we remove md support and close 10-20-30 issues, implementation for PHP7 support will not become easer.

Actually it will, and much more. There is a thing called cognitive overload: brain is able to consume certain amount of data, then it become very inefficient. Same happens here. We have 15 issues around Markdown, but there is no support for PHP 5.5 feature. Why? Because Markdown took all the energy we've got. And new issues are and keep popping. 

Trust me, I speak from experience. I've taken over maintenance of this package ~3 years ago, when it was 2 years dead and last version broken. There we so many features, that none was able to find and fix one line in the template generator.

All features were as wanted as Texy/Markdown/Parsedown..., but it didn't work.

At this state. This project don't work, and people are preffering Sami and PhpDocumentor, because they do.

**I see this as sinking ship. We are too heavy and we don't want to let go of great food we have and we ignore this huge whole in the deck.**

Let's make a compromise. I suggest we drop this for now and when we're done with PHP 7.1 feature and they work, we can put this back. Knowing we are safe, ship is fixed and we can enjoy those awesome meals.

jankal:
Yep @TomasVotruba I totally agree with this compromise.

@POPSuL is this fine for you?

jankal:
I won't wait for weeks until @POPSuL replies.

If checks pass as expected, I will merge this very soon (today or tomorrow).

TomasVotruba:
This has been merged to wrong branch. My fault.

I'll prepare new PR with event solution.

jankal:
Oh. I just assumed you had control over your `drop-*` branches. After short discussion I'd approve everything cause it's not master.

I don't know your plans cause I myself would have made just one branch to combine it all so that branch can then be merged into master. Now you have like five branches.

TomasVotruba:
I'm sorry for this mess. I'll be more careful now and since there is much less code know, it will be easier.

TomasVotruba:
@POPSuL It is now possible to modify docs output via event. 
You can easily access it via `EventSubscriber` and modify it to meet your needs - Markdown, Parsedown, Texy or anything you need.

jankal:
@TomasVotruba +1 for Events.

POPSuL:
Thanks for events!

PR #805, opened 2017-03-14
Use Symfony\EventDispatcher + add text event

TomasVotruba:
Few years ago I've added custom EventDispatcher to ApiGen. 

Now I trust this Symfony components to be stable and useful, so I suggest removing own approach and use Symfony one.

Replacing by it's integration to Nette: https://github.com/Symplify/SymfonyEventDispatcher#1-create-class-that-implements-symfonycomponenteventdispatchersubscriberinterface


jankal:
+1

jankal:
Don't we need tests for that? With tests 1957e4f3a28b7887c2f35669e6aca81ca081271f wouldn't have arised.

PR #806, opened 2017-03-14
drop deprecated

TomasVotruba:
substitute for #794, to master


TomasVotruba:
Already checked, mergin.

jankal:
+1

PR #813, opened 2017-03-22
drop magic elements

TomasVotruba:
tl;dr; Doesn't work properly anyway. Requires "standard magic". Promote best practice - OOP and explicit.

### The idea

I think we should promote best practice, clean code and modern features.
The code base is still huge. I was thinking about @method and @property, @property-read, @property-write magic annotations.

I also checked Sami (the other API doc generator) and they only support @property, not @method.
Also syntax for @property/@method may vary per user and so on. It would be difficult to make it correctly. Moreover if we still don't have PHP 5.5 features.

Also this ApiGen 5 aims on PHP 7.1+, so I guess this code would have more quality, thus this feature not so relevant. 

Back to the beginning, I consider this magic wrong - we should promote normal properties and methods instead.


TomasVotruba:
Ready to merge

TomasVotruba:
:+1:  Thanks

PR #823, opened 2017-03-29
drop ignoring @internal

TomasVotruba:
**This annotation clearly states that method is used for internal purpose** - usually public method, that should be private -, so it should not be displayed and turning it off doesn't make sense. It would be like using reflection to make public method private.

Show only in source code.


TomasVotruba:
Ready to merge

ek9:
Small comment regarding this on `@internal`:

> The @internal tag can be used as counterpart of the @api tag, indicating that the associated Structural Elements are used purely for the internal workings of this piece of software.
>
> An additional use of @internal is to add internal comments or additional description text inline to the Long Description. This may be done, for example, to withhold certain business-critical or confusing information when generating documentation from the source code of this piece of software.

Source: https://www.phpdoc.org/docs/latest/references/phpdoc/tags/internal.html

While from my understanding we do not support `@api` tag, perhaps this had use-cases from the previous users and we should consider bringing it back.

TomasVotruba:
Please try to keep features frozen until we're ready with 4-year old task of replacing Reflection.

We need to focus our attention there. Having that done, we can easily ad new features, don't worry.

PR #824, opened 2017-03-29
[WIP] Prefer config configuration

TomasVotruba:
Based on analysis of modern CLI PHP tools  - Sami & PHPStan, they don't have duplicated options in CLI and config. They only have most important options in Command configuration, other options are set via config.

**We should move this direction, also to allow adding services - like own filters, generators and event subscribers in the future.**

---

Other changes:

- drop Google opensearch, not API related nor used


jankal:
Just merged cause I liked the idea. Tests passed.  So for additional functionality please make new PR.

TomasVotruba:
Thank you

PR #826, opened 2017-03-30
drop description hiding + main option

TomasVotruba:
This is kind of non-intuitive feature, hides relevant content, broken on mobile and difficult to to use.
Sometimes I have to double click the method, to see anything, sometimes just once. Then it hides, pop-ups etc.

This PR will show everything that is relevant to the user.

---

As for main, I've checked few ApiGen output and none is using it. Also it's unusable for multiple sorting (logic step). 

I'd move attention to arguments `source` and `exclude`, where it makes more sense.

---

Also, since short description is not used anymore, move short and long description to simply `description`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,6 @@ To add another configuration, add `apigen.neon` to your root project.
 You can setup all these options:
 
 ```yaml
-main: ApiGen
 visibilityLevels: [public, protected] # array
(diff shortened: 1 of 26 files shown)


TomasVotruba:
Ready to merge

TomasVotruba:
Thank you @jankal

PR #835, opened 2017-04-03
Symfony 2.x & 3.x component support

ek9:
Add support for both 2.x and 3.x series of symfony components so ApiGen does not force specific version


ek9:
Fixes #835

PR #841, opened 2017-04-08
Cache dependencies between builds

olvlvl:
Cache dependencies between builds, so we can shave 30 sec from them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ language: php
 
 sudo: false
 
+cache:
+  directories:
+    - $COMPOSER_CACHE_DIR
+    - $HOME/.composer/cache
+
 matrix:
   include:
     - php: 7.1


TomasVotruba:
I'd rather not add this, since we have 1 build and it makes config less readable for others.

Having 5 builds, it might make sense.

olvlvl:
Cache is used by subsequent builds, not only builds of the same commit. e.g. https://travis-ci.org/Icybee/module-editor

TomasVotruba:
Either way, I didn't see much improvements before and after this PR.

olvlvl:
Well, you need at least two builds to see the benefit. The PR can only improve things, but do as you wish :)

TomasVotruba:
> Well, you need at least two builds to see the benefit.

Exactly.

> The PR can only improve things.

Project readability (1 year with few commits) is weak point of this I want to get rid off.

Thanks, I just wanted to explain my reasons.

PR #842, opened 2017-04-16
Copied Icanboogie theme

jankal:


jankal:
There seems to be some type of problem with CS-Fixer:
```
  [Symplify\EasyCodingStandard\Exception\Validator\CheckerIsNotSupportedException]  

  Checker "Symplify\CodingStandard\Sniffs\Classes\ClassDeclarationSniff" was        

  not found or is not supported. Use class that implements any of PHP_CodeSni       

  ffer\Sniffs\Sniff or PhpCsFixer\Fixer\FixerInterface.                             
```

TomasVotruba:
Based on many related changes not only in reflection, I've merged Default and Bootstrap theme to make it easies to maintain and accessible from mobile. 

Saying that, I'd say creating extra theme would be wise based on 5.x stable release, when the dust settles.

So closing this for now, but feel free to open on stable 5.x release, if you really won't like the design :+1:

jankal:
Sure. I'm watching you making "your thing" every day (my mail is full of GitHub).

PR #851, opened 2017-05-04
Change visibilityLevels to accessLevels

### LINKED ISSUE #848: Not all functions being found for class.
## Description

Generation does not seem to be picking up all of the functions for classes.

Preparing documentation site for the following project: 
https://github.com/8fold/eventbrite-sdk-php

Focusing on the following file:
https://github.com/8fold/eventbrite-sdk-php/blob/develop/src/Eventbrite.php

When I generate the documentation I only see the long description and the `startEventbrite` method. Nothing else really from itself - near as I can tell.

What am I doing wrong here?

screen shot 2017-04-30 at 9 49 02 pm

joshbruce:
Fixes #848

After changing the key name to match the flag in the CLI, all seems
right with the world.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ the root of your project:
 ```yaml
 source: [src]           # directory(-ies) to scan PHP files from
 destination: docs       # destination directory to generate API docs in
-visibilityLevels: [public, protected] # array
+accessLevels: [public, protected] # array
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Thank you for PR.

This feature will be resolved via own service in ApiGen 5.0, so I'm closing this.

PR #875, opened 2017-06-05
[Generator] add reflection lists

TomasVotruba:
Inspired by Sami

image


packages/Reflection/src/Reflection/Class_/ClassReflection.php
```diff
@@ -128,12 +128,13 @@ public function getInheritedDescription(): ?string
...
     {
-        if (! $this->betterClassReflection->getParentClass()) {
+        $parentClassName = get_parent_class($this->getName());
+        if ($parentClassName === false) {
             return null;
         }
 
         return $this->transformerCollector->transformSingle(
-            $this->betterClassReflection->getParentClass()
+            ReflectionClass::createFromName($parentClassName)
```
TomasVotruba: Note: much faster according to Blackfire

PR #876, opened 2017-06-05
[Element] Namespace collector + [Theme] navbar

TomasVotruba:
## Breadcrums added

So you know, where you are and how to get up!

### Before

before

### After

after

image

image


f3l1x:
I like it.

TomasVotruba:
Thank you!

PR #880, opened 2017-06-07
[DI] use Symfony and PSR-11 container

TomasVotruba:
Thanks for the idea @mhujer

I need bit help with this. Looks like only classes can be autowired, not interfaces: https://travis-ci.org/ApiGen/ApiGen/builds/240583095#L302 Which makes me sad, because breaks inversion of control.

Might be caused by this: https://github.com/symfony/symfony/pull/22295

Is that right, or am I doing something wrong?

---

## Why?

- This will open-up the whole world to work with the code.
- It makes advantages of PSR-11 container - simple to tune or replace.
- It makes it easier to develop and add new services thanks to [Symfony 3.3+ DI features](https://www.tomasvotruba.cz/blog/2017/05/07/how-to-refactor-to-new-dependency-injection-features-in-symfony-3-3/)


TomasVotruba:
It might be related also to https://github.com/symfony/symfony/pull/22254, which makes more sense.

Issue #888, opened 2017-06-15
How to run ApiGen (on Windows System)?


brunoaugusto:
I've been away of development for a very long time and since I came back I've been updating a big system I have, mostly for personal uses, and today I decided to generate its documentation.

However, I develop under Windows system. And don't fool yourselves, this doesn't make me a worse developer than a Linux user at all. In past, not sure which version but I can tell it was before ApiGen become a Composer resource (or whatever it's called) I could simply download the branch here, access the extracted folder directly through command prompt and generate the documentation by only adjusting the paths.

The documentation was generated as best as it could. If there was anything unexpected, missing or wrong with the source code parsed, a log file was generated and I cold quickly fix.

Well, I gave it a shot this new way. I installed Composer, ApiGen and tried to run. First thing I've got:

> The filename, directory name or volume label syntax is incorrect

I tried a few different path formats without success. Then I moved my command pointer directly to one level above my source directory and ran the command through there, **following the Readme.md instructions**;

> D:\bin\vendor\bin\apigen generate src myFolder --destination docs

Both directories exist and the output:

> [Symfony\Component\Console\Exception\RuntimeException]
> Too many arguments, expects arguments "command"

But now with a list of valid configuration options.

With the output list I saw that **src** is not valid. That I should use `--source myFolder` or at least `-s myFolder`

Once again, by trial and error, I adjusted and finally I saw the expected; **Scanning sources and parsing** only to receive another error few seconds later:

> Unexpected token found.

But what token? Any details about this error? Where's the log? I looked at Composer install directory, ApiGen Composer required directory, my output fodlers... Nope, *nada*

I understand how parsing works. It's really a pain and if anything goes wrong the process blows up. But differently of the very, very, very old versions, I'm sorry to say this bluntly, but it seems to be lacking some "care" with these little details.

TomasVotruba:
Hi @brunoaugusto,

thanks for detailed and very clear report. I think we can make this work.

First: are you running the version as README is? In this case `dev-master`?

**We focus on master and 5.x release only at the moment, so try it with that please.**

How to try master you'll find here: #886 (comment)

brunoaugusto:
I installed as described in the README.md, there wasn't any mention to a `dev-master`.

Well, I tried a few more things. First I changed the PHP version Composer was using. PHP 7 is "too new" and even though we can install any version we want on dedicated servers, that's not the reality in my country, but that's beside the point.

[...]

PR #892, opened 2017-06-20
Expand Composer AutoLoader Searchable Paths

brunoaugusto:
Expanding the the searchable path to locate Composer's AutoLoader in order to, possibly, Fix issue #888


TomasVotruba:
Thank you for your first PR!

I've fixed coding standard and used more common function for this.

PR #895, opened 2017-06-22
 implemented TraitMethodReflection::isDeprecated() [fixes #854] 

### LINKED ISSUE #854: "minimal example configuration" in readme.md has bug
Array notation is required for `exclude`, see below:

`exclude: tests`

Causes:

```Argument 2 passed to ApiGen\Scanner\Scanner::scan() must be of the type array, string given, called in vendor/apigen/apigen/src/Command/GenerateCommand.php on line 165 and defined...```

Needs to be like this:

`exclude: [tests]`

### LINKED ISSUE #894: Reflection: Unimplemented methods
I've found another problem.

In `ApiGen\Reflection\Reflection\Trait_\TraitMethodReflection` there are unimplemented methods. Particularily, I've got this exception:

```
[ApiGen\Exception\Templating\FailedRenderFileException]                      
Rendering of "......./vendor/apigen/apigen/packages/ModularConfiguration/src/Option/../../../../packages/ThemeDefault/src/class.latte" template to "...../class-test.html" file failed due to: Return value of ApiGen\Reflection\Reflection\Trait_\TraitMethodReflection::isDeprecated() must be of the type boolean, none returned 
```

It is logic for the function `TraitMethodReflection::isDeprecated()` has strict typehint. But, what is strange, I am getting this error only when I am generating docs by running `apigen generate`. When I run `phpunit`, it generates documentation without fail. I even tried to edit file `/packages/ThemeDefault/src/class.latte` to see whether the condition `n:class="$ownTrait->isDeprecated() ? deprecated"` is not skipped and it wasn't.

So, where problem lies is obvious: in unimplemented methods.

But why test passes? It seems to me like the strict typing is ignored somehow. I do not know :)

vlastavesely:
I've fixed #894

### DIFF
--- a/packages/Reflection/src/Reflection/Trait_/TraitMethodReflection.php
+++ b/packages/Reflection/src/Reflection/Trait_/TraitMethodReflection.php
@@ -93,7 +93,12 @@ final class TraitMethodReflection implements TraitMethodReflectionInterface, Tra
 
     public function isDeprecated(): bool
     {
-        // TODO: Implement isDeprecated() method.
+        if ($this->betterMethodReflection->isDeprecated()) {
+            return true;
+        }
+
+        return $this->getDeclaringTrait()
+            ->isDeprecated();
     }
 
     public function getDescription(): string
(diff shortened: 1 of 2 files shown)


vlastavesely:
Do you know any good variable dumper that could be used for dumping non-scalar variables?

TomasVotruba:
Great job, thank you!

PR #897, opened 2017-06-23
updated test [failing]

vlastavesely:
`ClassMethodReflection::getDefaultValueDefinition()` can return mixed types but typehint is only for string. The same goes for `TraitMethodReflection::getDefaultValueDefinition()`.

Some variable dumper must be added or real definition loaded in some way.

Any tips how to solve it?

### DIFF
--- a/packages/Element/src/Latte/Filter/PhpHighlightFilter.php
+++ b/packages/Element/src/Latte/Filter/PhpHighlightFilter.php
@@ -1,0 +1,32 @@
+<?php declare(strict_types=1);
+
+namespace ApiGen\Element\Latte\Filter;
+
+use ApiGen\Contract\Templating\FilterProviderInterface;
+use ApiGen\SourceCodeHighlighter\SourceCodeHighlighter;
+
+final class PhpHighlightFilter implements FilterProviderInterface
+{
+    /**
+     * @var SourceCodeHighlighter
+     */
+    private $highlighter;
+
+    public function __construct(SourceCodeHighlighter $highlighter)
+    {
+        $this->highlighter = $highlighter;
+    }
+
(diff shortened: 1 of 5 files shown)


TomasVotruba:
Thanks for propper test case.

I think that solution is to convert this to plain string and return it
E.g. 

```php
public $arrayProperty = ['cat', 'dog'];
```

return:

```
" = ['cat', 'dog'];"
```

Probably using some function like [`htmlspecialchars`](http://php.net/manual/en/function.htmlspecialchars.php). Not sure exactly which now.

vlastavesely:
> I think that solution is to convert this to plain string and return it.

Oh yes, it is. But there are two possible solutions. One is to extract somehow the `" = ['cat', 'dog'];"` string from file directly or use output of `betterPropertyReflection->getDefaultValue()` and pas it into some dumper like `Nette\PhpGenerator\Helpers::dump()` or `var_export()`. I am not sure what would be better.

TomasVotruba:
First one would be better to comply with currenct ApiGen architecture. I'm not sure how much work that is.
Feel free to pick one that suits you best and is easy to add.

vlastavesely:
I've looked on that and first solution does not seem to be easy. Probably, I try the second one. It can be changed later.

TomasVotruba:
I've rebased the PR on `master`, as there were some conflicts.

vlastavesely:
Fixed. In the future, the better dumper can be used but right now, we need it working in any way.

Now it looks like this:
screenshot

TomasVotruba:
I like it, pretty neat solution! :+1: 

Thanks for the screenshots, it saves me lot of time running tests and generating own output.

Issue #884, opened 2017-06-10
[5.x][help needed] Fix source code highlighting javascript and design


TomasVotruba:
## Required skills

- bit CSS
- Javascript

## How it Should Work

When you click to source code on specific line, **it should highlight** (same as on Github). See [Doctrine API](http://www.doctrine-project.org/api/orm/2.5/source-class-Doctrine.ORM.Cache.Logging.StatisticsCacheLogger.html#50-58) or just watch me recoding it for you:

![highlight](https://user-images.githubusercontent.com/924196/27002266-9d91d6b8-4ddd-11e7-9c87-0a58f3b165ba.gif)

I've upgrade to new Jquery 3.1, dropped lof of Javascript and broke it. I'm no Javascript master, so need help with this.

And it needs to improve code style as before, now it's influenced by Twitter Bootstrap:

![image](https://user-images.githubusercontent.com/924196/27002337-baf74e66-4ddf-11e7-8d31-c336e12bd5df.png)

## Important files with source code

- [main.js section for highlighting code](https://github.com/ApiGen/ApiGen/blob/4a8a08bc165dd09504af1c1db61d28a34c712514/packages/ThemeDefault/src/resources/main.js#L73-L90)
- [main.js section for saving selected lines](https://github.com/ApiGen/ApiGen/blob/4a8a08bc165dd09504af1c1db61d28a34c712514/packages/ThemeDefault/src/resources/main.js#L92-L158) - not sure what is this for, maybe to keep code highlighted after page refresh

Any help with this appreciated. **Thanks to you ApiGen 5.x can be here faster :+1:**

vlastavesely:
I have it almost done. Probably today I am gonna send a PR.

TomasVotruba:
@vlastavesely Thanks for taking this. That would be amazing.

TomasVotruba:
Closed by #898 

Thank you @vlastavesely

PR #898, opened 2017-06-26
fixed souce code line highlighting

vlastavesely:
Related to #884 

My goal was to imitate behaviour of latest Nette's API and one of Doctrine ORM. There was a little problem with line numbers styling for in former versions of ApiGen number and source code has been wrapped in one HTML element and in current version this is done differently. But I solve it.

Autocomplete has been removed temporarily, for it broke the highlighting.

### DIFF
--- a/packages/ThemeDefault/src/@layout.latte
+++ b/packages/ThemeDefault/src/@layout.latte
@@ -39,11 +39,11 @@
         </div>
 
         {* jquery autocomplete *}
-        <script src="https://code.jquery.com/jquery-3.2.1.min.js" defer></script>
-        <script src="https://ajax.googleapis.com/ajax/libs/jqueryui/1.12.1/jquery-ui.min.js" defer></script>
+        <script src="https://code.jquery.com/jquery-3.2.1.min.js"></script>
+        <script src="https://ajax.googleapis.com/ajax/libs/jqueryui/1.12.1/jquery-ui.min.js"></script>
 
-        <script src="/resources/main.js"></script>
-        <script src="/elementlist.js"></script>
+        <script src="elementlist.js"></script>
+        <script src="resources/main.js"></script>
     </body>
 </html>
 
--- a/packages/ThemeDefault/src/resources/main.js
+++ b/packages/ThemeDefault/src/resources/main.js
@@ -1,9 +1,9 @@
 // $( function() {
-$(window).load(function() {
+$(document).ready(function() {
     var $document = $(document);
 
     // Content
-
+/*
     var availableTags = [
         "ActionScript",
         "Able"
@@ -69,6 +69,8 @@ $(window).load(function() {
                 }
                 return !autocompleteFound && '' !== $('#search input[name=cx]').val();
             });
+*/
+
 
     // Select selected lines
     var matches = window.location.hash.substr(1).match(/^\d+(?:-\d+)?(?:,\d+(?:-\d+)?)*$/);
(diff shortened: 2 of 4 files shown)


vlastavesely:
Tested in FF and Chromium.

TomasVotruba:
@vlastavesely Thanks a lot for the PR.

Could you please add before/after screenshot? This is related to the design and hard to see from the code.

vlastavesely:
Now it looks like this:
screenshot


And for comparison your own screenshot how it used to look:
old

packages/ThemeDefault/src/@layout.latte
```diff
@@ -39,11 +39,11 @@
         </div>
 
         {* jquery autocomplete *}
-        <script src="https://code.jquery.com/jquery-3.2.1.min.js" defer></script>
-        <script src="https://ajax.googleapis.com/ajax/libs/jqueryui/1.12.1/jquery-ui.min.js" defer></script>
+        <script src="https://code.jquery.com/jquery-3.2.1.min.js"></script>
+        <script src="https://ajax.googleapis.com/ajax/libs/jqueryui/1.12.1/jquery-ui.min.js"></script>
```
TomasVotruba: What was the motivation to remove this?

packages/ThemeDefault/src/@layout.latte
```diff
@@ -39,11 +39,11 @@
...
 
-        <script src="/resources/main.js"></script>
-        <script src="/elementlist.js"></script>
+        <script src="elementlist.js"></script>
+        <script src="resources/main.js"></script>
```
TomasVotruba: I recall there was some issue for path without `/`. What was the motivation to remove and switch order?

vlastavesely: switching: main.js  is going to use ApiGen.elements that has to be declared before main.js is loaded.
when path was with `/`, browser failed to load the files, for they are not really in filesystem root. Do you use Linux?

packages/ThemeDefault/src/@layout.latte
vlastavesely: If I understand it correctly, this parameter forces loaded JS to be executed after page is parsed. The next two JS files, however were loaded in a standard way and so executed before jQuery has been loaded. I has problems with missing `$`.

packages/ThemeDefault/src/@layout.latte
TomasVotruba: 1. I see.

2. Yes I do. I tried it locally and it works.

packages/ThemeDefault/src/@layout.latte
TomasVotruba: I see, ok. I recall I had similar issue. This was just experimental feature for me, I don't really use it very often nor know a lot a about it.

packages/ThemeDefault/src/resources/main.js
```diff
@@ -146,7 +146,7 @@ $(document).ready(function() {
             }
         }
 
-        hash = hash.join(',');
+        hash = '#' + hash.join(',');
```
TomasVotruba: ### Before

before


### After

after

TomasVotruba:
I've added a small fix, it's ready to merge otherwise.

Tested in Chrome on Linuxu (Ubuntu 17.04)

Can I merge it?

vlastavesely:
Yes, you can.

Your screenshots are giving us an answer to the problem with `/` in path :) You use HTTP server, where its root is exact the same with root of the generated docs. I do open the files via `file://` protocol and so, their path is never the same with filesystem's root. That is the reason why it worked for you but not for me.

TomasVotruba:
Yep, thats the case. I think subdirectory like `localhost/api/` would fail too.

Thanks for your work :bowing_man:

PR #903, opened 2017-06-28
fixed function parameter reflection

vlastavesely:
```
ApiGen\Exception\Templating\FailedRenderFileException: Rendering of "/home/*/apigen/packages/ThemeDefault/src/function.latte" template to "/tmp/_apigen_cache/function-ApiGen.Tests.Generator.Source.someOtherFunction.html" file failed due to: Return value of ApiGen\Reflection\Reflection\AbstractParameterReflection::getAnnotation() must be an instance of phpDocumentor\Reflection\DocBlock\Tags\Param or null, instance of phpDocumentor\Reflection\DocBlock\Tags\Return_ returned
```

```
ApiGen\Exception\Templating\FailedRenderFileException: Rendering of "/home/*/apigen/packages/ThemeDefault/src/function.latte" template to "/tmp/_apigen_cache/function-ApiGen.Tests.Generator.Source.someOtherFunction.html" file failed due to: Call to undefined method ApiGen\Reflection\Reflection\Function_\FunctionParameterReflection::isDefaultValueAvailable()
```

```
ApiGen\Exception\Templating\FailedRenderFileException: Rendering of "/home/*/apigen/packages/ThemeDefault/src/function.latte" template to "/tmp/_apigen_cache/function-ApiGen.Tests.Generator.Source.someOtherFunction.html" file failed due to: Argument 1 passed to ApiGen\Annotation\Latte\Filter\AnnotationFilterProvider::ApiGen\Annotation\Latte\Filter\{closure}() must implement interface ApiGen\Reflection\Contract\Reflection\Partial\AnnotationsInterface, instance of ApiGen\Reflection\Reflection\Function_\FunctionParameterReflection given, called in /tmp/_latte_cache/ThemeDefault-src-function.latte--2096778438.php on line 154
```

### DIFF
--- a/packages/Reflection/src/Reflection/AbstractParameterReflection.php
+++ b/packages/Reflection/src/Reflection/AbstractParameterReflection.php
@@ -2,6 +2,7 @@
 
 namespace ApiGen\Reflection\Reflection;
 
+use ApiGen\Annotation\AnnotationList;
 use ApiGen\Reflection\Contract\Reflection\AbstractParameterReflectionInterface;
 use ApiGen\Reflection\Contract\Reflection\Class_\ClassMethodReflectionInterface;
 use ApiGen\Reflection\Contract\Reflection\Class_\ClassReflectionInterface;
@@ -67,6 +68,11 @@ abstract class AbstractParameterReflection implements AbstractParameterReflectio
         return $classOrInterfaceReflection;
     }
 
+    public function isDefaultValueAvailable(): bool
+    {
+        return $this->betterParameterReflection->isDefaultValueAvailable();
+    }
+
     public function getDefaultValueDefinition(): ?string
     {
         if ($this->betterParameterReflection->isDefaultValueAvailable()) {
@@ -118,7 +124,7 @@ abstract class AbstractParameterReflection implements AbstractParameterReflectio
     private function getAnnotation(): ?Param
     {
         $declaringReflection = $this->getDeclaringReflection();
-        $annotations = $declaringReflection->getAnnotations();
+        $annotations = $declaringReflection->getAnnotation(AnnotationList::PARAM);
 
         if (empty($annotations[$this->betterParameterReflection->getPosition()])) {
             return null;
--- a/packages/ThemeDefault/src/function.latte
+++ b/packages/ThemeDefault/src/function.latte
@@ -51,7 +51,7 @@
                     <span class="property-name">{if $parameter->isPassedByReference()}&amp; {/if}${$parameter->getName()}</span>
                     {if $parameter->isDefaultValueAvailable()} = {$parameter->getDefaultValueDefinition()|highlightValue:$function|noescape}{elseif $parameter->isVariadic()},…{/if}
                 {/block}</code></td>
-                <td>{$parameter|description}</td>
+                <td>{if $parameter instanceof \ApiGen\Reflection\Contract\Reflection\Partial\AnnotationsInterface}{$parameter|description}{/if}</td>
             </tr>
         </table>
     {/if}
--- a/tests/Generator/Source/SomeFunction.php
+++ b/tests/Generator/Source/SomeFunction.php
@@ -6,6 +6,10 @@ function someFunction(): void
 {
 }
 
-function someOtherFunction(): void
+/**
+ * Do not add param annotations here!
+ * @return void
+ */
+function someOtherFunction($paramWithoutTypeHint)
 {
 }
--- a/packages/Reflection/src/Contract/Reflection/AbstractParameterReflectionInterface.php
+++ b/packages/Reflection/src/Contract/Reflection/AbstractParameterReflectionInterface.php
@@ -14,6 +14,8 @@ interface AbstractParameterReflectionInterface extends AbstractReflectionInterfa
      */
     public function getTypeHintClassOrInterfaceReflection();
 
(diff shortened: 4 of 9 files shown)


TomasVotruba:
Thanks for the PR. Could you please more description? It's difficult for me to read your intention based on 3 errors.

packages/ThemeDefault/src/function.latte
```diff
@@ -51,7 +51,7 @@
                     <span class="property-name">{if $parameter->isPassedByReference()}&amp; {/if}${$parameter->getName()}</span>
                     {if $parameter->isDefaultValueAvailable()} = {$parameter->getDefaultValueDefinition()|highlightValue:$function|noescape}{elseif $parameter->isVariadic()},…{/if}
                 {/block}</code></td>
-                <td>{$parameter|description}</td>
+                <td>{$parameter->getDescription()}</td>
```
TomasVotruba: Why is this needed? It would disable possibility to tune description via event. That's why there is filter "description" here, that calls `->getDescription()` internally.

tests/Generator/Source/SomeFunction.php
```diff
@@ -6,6 +6,9 @@ function someFunction(): void
 {
 }
 
-function someOtherFunction(): void
+/**
+ * @return void
+ */
+function someOtherFunction($paramWithoutTypeHint): void
```
TomasVotruba: Could you keep both types of functions - returning `: void`, one with annotation, second without?

So both cases are covered with test.

packages/Reflection/src/Reflection/AbstractParameterReflection.php
```diff
@@ -67,6 +67,11 @@ public function getTypeHintClassOrInterfaceReflection()
         return $classOrInterfaceReflection;
     }
 
+    public function isDefaultValueAvailable(): bool
```
TomasVotruba: Could please add this method to `ParameterReflectionInterface`?

vlastavesely: you mean `AbstractParameterReflectionInterface`?

packages/ThemeDefault/src/function.latte
vlastavesely: yes, it is mistake. in the evening I watched some of your commits and realized that this is not right. I am going to edit it.

Where latte filters are defined?

vlastavesely:
It could happen that a parameter of a function does not have anotation. If if happens, there is not returned instance of `AnnotationsInterface` and latte filter fails because of it.

Separated problem was missing `AbstractParameterReflection::isDefaultValueAvailable()`.

And third problem was that function `AbstractParameterReflection::getAnnotation()` could return annotation for `@return` instead `@param`.

packages/Reflection/src/Reflection/AbstractParameterReflection.php
TomasVotruba: Yep

TomasVotruba:
I see, now I get it. Thanks for expalining!

TomasVotruba:
As for `Return`/`Param` annotation, maybe (there should be)[$annotations = $declaringReflection->getAnnotations()]:

```php
$annotations = $declaringReflection->getAnnotation(AnnotationList::PARAM);
```

That should prevent both `instanceof Param` checks bellow and in a template.

vlastavesely:
yes, this is better. I was not aware of fact that function getAnnotations() can take an argument. :)

TomasVotruba:
@vlastavesely Good point. Maybe `getAnnotationsByType()` would be better.

vlastavesely:
In template, problem persists:
```
ApiGen\Exception\Templating\FailedRenderFileException: Rendering of "/home/*/apigen/packages/ThemeDefault/src/function.latte" template to "/tmp/_apigen_cache/function-ApiGen.Tests.Generator.Source.someOtherFunction.html" file failed due to: Argument 1 passed to ApiGen\Annotation\Latte\Filter\AnnotationFilterProvider::ApiGen\Annotation\Latte\Filter\{closure}() must implement interface ApiGen\Reflection\Contract\Reflection\Partial\AnnotationsInterface, instance of ApiGen\Reflection\Reflection\Function_\FunctionParameterReflection given, called in /tmp/_latte_cache/ThemeDefault-src-function.latte--2096778438.php on line 154
```

TomasVotruba:
I see. Let's have it this way for now. Is it ready to merge?

vlastavesely:
Only thing that remains, would be method in class and in trait - I am afraid that it could cause same problem. Give me a few minutes.

TomasVotruba:
Allright, no need to hurry, I was just curious. Take your time

vlastavesely:
I tried it and it was a little bit buggy.

Main new thing: in `method.latte` there are information about overriden method and what interfaces' method is implemented. But trait cannot extend or implement anything. So now it is shown just or class.

TomasVotruba:
I'll look on later.

Could you fix cs as shown in Travis?

vlastavesely:
I already fixed. And then realized that my PR is dumplicate to #909 :)

TomasVotruba:
I see, great. You're so fast I can't keep up on my mobile phone :)

TomasVotruba:
I see, great. You're so fast I can't keep up on my mobile phone :)

TomasVotruba:
Thank you for the the fix and compleeting missing methods!

Issue #902, opened 2017-06-28
Messy console output


vlastavesely:
When generating API, a progressbar is shown. It works until it reaches end:

```
$ gen.sh
 11/27 [===========>----------------]  40% 8 secs/20 secs 14.0 MiB
```

But after 100% is reached, a few more lines are printed so all output looks like that:

```
$ gen.sh
 27/27 [============================] 100% 31 secs/31 secs 14.0 MiB - Finished!
 28/28 [============================] 100% 31 secs/31 secs 14.0 MiB - Finished!
 29/29 [============================] 100% 31 secs/31 secs 14.0 MiB - Finished!
 30/30 [============================] 100% 31 secs/31 secs 14.0 MiB - Finished!
 31/31 [============================] 100% 31 secs/31 secs 14.0 MiB - Finished!
 32/32 [============================] 100% 31 secs/31 secs 14.0 MiB - Finished!
```

It does not seem to be right behavior. Or are these some postprocess steps?

TomasVotruba:
That looks broken. Could you look to `ProgressBar` class?
It seems like it advances too much: https://github.com/ApiGen/ApiGen/blob/master/src/Console/Helper/ProgressBar.php#L43

TomasVotruba:
Closed by #905

nelson6e65:
The maximum counter is failing, but I didn't find out how to tests that. :sweat_smile:

So, I made a #922 that puts that check in destructor of progress bar, so it will not going to break the progress bar. 

In self apigen api generation, is set to 110, but the counter is incremented 1 by 1 until 115. For example: https://travis-ci.org/ApiGen/ApiGen/builds/249262068#L725

PR #905, opened 2017-06-29
Fixed messy output in console.

vlastavesely:
In short, `StepCounter` did not count files like `index.html`, `classes.html` etc. Beacuse of this, progressbar's maximum was lesser than actual count of generated files.

It solves #902


TomasVotruba:
I suspected some numbers didn't match, glad you found it.

Thank you.

Issue #900, opened 2017-06-28
PHP Fatal error:  Uncaught Error: Class '...\ContainerFactory' not found in .../vendor/apigen/apigen/bin/apigen:13


nelson6e65:
Hi. I'm trying to install Apigen `v5.0.0-RC1` but is not working. 

I get this error in https://travis-ci.org/nelson6e65/php_nml/jobs/247806391#L320:

`PHP Fatal error:  Uncaught Error: Class 'ApiGen\DependencyInjection\Container\ContainerFactory' not found in ...vendor/apigen/apigen/bin/apigen:13`

https://github.com/ApiGen/ApiGen/blob/v5.0.0-RC1/bin/apigen#L13

**Edit**
The problem ***is not*** on ApiGen install, but in `apigen generate`.

nelson6e65:
Using `dev-master` I'm still getting this error.

https://travis-ci.org/nelson6e65/php_nml/jobs/247810847#L314

`Fatal error: Uncaught Error: Class 'ApiGen\DependencyInjection\Container\ContainerFactory' not found in /home/travis/.composer/vendor/apigen/apigen/bin/apigen on line 23`

TomasVotruba:
Hey @nelson6e65 , thanks for reporting the issue.

How did you install this package?

nelson6e65:
Hi. I installed with Composer, as global package: **`composer global require apigen/apigen:dev-master`** (https://github.com/nelson6e65/php_nml/blob/master/script/travis-ci/deploy-documentation#L10)

> Note: I have not PHP 7.1 installed in my computer, so I used Travis CI, because ApiGen is not working on my machine (PHP 5.6) to perform `generate` locally.

TomasVotruba:
Hey, the link on Travis is leading to newer version where is no error. Could you provide exact link please?

Instead of `composer global` I recommend using `composer create-project`.

You can also try to modify directory loading in `bin\apigen`: https://github.com/ApiGen/ApiGen/blob/master/bin/apigen#L10
Not sure how on Travis.

Why don't you install PHP 7.1 instead?

nelson6e65:
> Hey, the link on Travis is leading to newer version where is no error. Could you provide exact link please?

Oh, sorry... Incorrect line line for dev-master:
https://travis-ci.org/nelson6e65/php_nml/jobs/247810847#L320


> Why don't you install PHP 7.1 instead?

[...]

PR #908, opened 2017-06-30
[testing] Avoid using autoload file from current working dir

nelson6e65:
Test if works what I mention in https://github.com/ApiGen/ApiGen/issues/900#issuecomment-312199321.

Avoid load the project vendors instead of ApiGen vendor.

> Sorry, I have not PHP 7.1 in my computer to test this locally.

### DIFF
--- a/bin/apigen
+++ b/bin/apigen
@@ -8,7 +8,6 @@ use Symfony\Component\Console\Application;
 gc_disable();
 
 $possibleAutoloadPaths = [
-    getcwd() . '/vendor/autoload.php',
     __DIR__ . '/../vendor/autoload.php',
     __DIR__ . '/../../../autoload.php'
 ];


nelson6e65:
Well... I think fa coding standard fails since #905: 
https://travis-ci.org/ApiGen/ApiGen/builds/248358601#L323

> See #909

nelson6e65:
This can, actually, improved more, because switch between `'/../vendor/autoload.php'` and `'/../../../autoload.php'` is that the first will be used for *global install* (or local development) and the other is used if apigen is a *dev dependency*: **`apigen generate -h`** (*and* `bin/apigen generate -h`) or **`vendor/bin/apigen generate -h`** usage, respectively.

I'm missing something?

TomasVotruba:
Great work! I think you didn't miss anything

Issue #901, opened 2017-06-28
Fails to generate some classes without throwing error


vlastavesely:
I tried generate API for `Nette\Neon` and API for class `Decoder` is not generated although no error is produced. In namespace summary `namespace-Nette.Neon.html` there is a link to `class-Nette.Neon.Decoder.html` but file is missing. I have no idea why it is failing...

Tested on `Nette\Neon` 2.4-20170517.

TomasVotruba:
Could you send a simple test case please?

vlastavesely:
Oh, I think I found what it is causing. `Decored` is marked as `@internal`.
Internal classes should be omitted in class list, should not?

TomasVotruba:
Yes, they should. Where exactly is that failing in template?

vlastavesely:
This is problem of `NamespaceReflectionCollector`, not of a template.I will fix it tomorrow.

TomasVotruba:
Ok, thank you.

Btw, if you have any idea how to improve template exception to give more relevant information, we can go for it. I realize they could be more verbose and helpful.

vlastavesely:
I have no idea :) It can fail for so many different exceptions so it would not be possible...

TomasVotruba:
Fixed by #910 and #917

PR #910, opened 2017-06-30
internal classes removed from namespace summary

vlastavesely:
fix for #901 

I think this does not need more description ;)

### DIFF
--- a/packages/Annotation/src/AnnotationList.php
+++ b/packages/Annotation/src/AnnotationList.php
@@ -12,6 +12,11 @@ final class AnnotationList
     /**
      * @var string
      */
+    public const INTERNAL = 'internal';
+
+    /**
+     * @var string
+     */
     public const PARAM = 'param';
 
     /**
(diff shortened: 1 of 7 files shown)


TomasVotruba:
Thanks for the PR. Link to issue is just fine :+1: 

Could you add test to get this coverred?
Actually just add a class with `@internal` annotation here: https://github.com/ApiGen/ApiGen/tree/master/packages/Element/tests/ReflectionCollector/NamespaceReflectionCollectorSource

There might be an easier way (different bug) to do this. There should be no internal reflections anywhere.

TomasVotruba:
Thanks, I'll check it tonight.

TomasVotruba:
Could you please your name your branches different than "master"?
That would allow me to fetch them locally easily.

```bash
git fetch origin pull/ID/head:BRANCHNAME
```

Now it fails on:

image

TomasVotruba:
Replaced by #917

Issue #912, opened 2017-06-30
Is there possibility to generate API of another repository?


vlastavesely:
Basically, I have directories like `/git/vendor/repo` and inside a PHP library in `src` dir. And what I need is to generate API for all repositories together. How to solve this use case?

If I create a directory somewhere and install ApigGen via composer with the following `composer.json`:

```
{
	"require": {
		"apigen/apigen": "5.0.0-RC1",
	},
	"minimum-stability": "dev",
	"autoload": {
		"classmap": ["/git/myvendor/localization/src", "...."]
	},
	"scripts": {
		"apigen": [
			"rm -rf ./docs/*",
			"vendor/bin/apigen generate /git/myvendor/localization/src /git/myvendor/..... --destination docs"
		]
	}
}
```

it fails for dependencies are not autoloaded or are different from the dependencies used by ApiGen (typically `nette/utils` - my library needs `3.0`, ApiGen uses `2.4`).

Any tips here?

TomasVotruba:
For these use cases, use `composer create-project ...` approach.

Here is example of how to: https://github.com/nette/application/pull/182/commits/b6cb107128534012577f8d71b22d37bc3a0173fb

vlastavesely:
And what about autoload of classes outside the source?

I just tested it in a way you suggested but it failed on my class extending `Nette\Security\User` that could not be autoloaded for it is not in `src` directory from which I generate the API.

```
  [ApiGen\Exception\Templating\FailedRenderFileException]                      
  Rendering of "/home/*/api/apigen/packages/ModuarConfiguration/src/Option/../../../../packages/ThemeDefault/src/class.latte" template to "/home/*/api/docs/class-Andromeda.Authorization.User.html" file failed due to: Roave\BetterReflection\Reflection\ReflectionClass "Nette\Security\User" could not be found in the located source   
```

TomasVotruba:
Interesting. This might be caused by missing locator for `BetterReflection`. Could send a test case for class i `/src` with missing class in it? I'll investigate.

vlastavesely:
1) installed Apigen
```
composer create-project apigen/apigen -s dev
```

2) placed this  into `src`

```
<?php

namespace Test;

use Nette;

class User extends Nette\Security\User
{

}
```

3) run
```
apigen/bin/apigen generate src --destination docs
```

4) Get the exception

TomasVotruba:
Could you send that to PR?

TomasVotruba:
[...]

PR #915, opened 2017-06-30
tests: added class extending from another project

vlastavesely:
Test for #912

But from `phpunit`it ends with FATAL ERROR instead of exception. It confuses me a little - is it right or have I miss something...?

### DIFF
--- a/tests/Generator/ExtendingClassGeneratorTest.php
+++ b/tests/Generator/ExtendingClassGeneratorTest.php
@@ -1,0 +1,36 @@
+<?php declare(strict_types=1);
+
+namespace ApiGen\Tests\Generator;
+
+use ApiGen\Generator\ClassGenerator;
+use ApiGen\Reflection\Parser\Parser;
+use ApiGen\Tests\AbstractContainerAwareTestCase;
+
+final class ExtendingClassGeneratorTest extends AbstractContainerAwareTestCase
+{
+    /**
+     * @var ClassGenerator
+     */
+    private $classElementGenerator;
+
+    protected function setUp(): void
+    {
+        /** @var Parser $parser */
+        $parser = $this->container->get(Parser::class);
+        $parser->parseDirectories([__DIR__ . '/ExtendingSources']);
+
+        $this->classElementGenerator = $this->container->get(ClassGenerator::class);
+    }
+
+    public function testGenerate(): void
+    {
+        $this->classElementGenerator->generate();
+
+        $this->assertFileExists(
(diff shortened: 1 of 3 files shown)


TomasVotruba:
I think `BetterReflection` requires the class to be loaded by design.

I'm not sure it would solved your issue, but could your add the parent class to another directory (not the ApiGen generates it from)? So we see, if the class can be autoloaded.

---

As for different vendor issue, do you have any idea for solution?

I think we can use [`ComposerSourceLoader`](https://github.com/Roave/BetterReflection/blob/master/src/SourceLocator/Type/ComposerSourceLocator.php#L17) to [our Parser](https://github.com/ApiGen/ApiGen/blob/b13e5bc7b59c7af4b13f34e0dea76bf0b68377c2/packages/Reflection/src/Parser/Parser.php#L120), but we'd need to pass location to the `vendor/autoload.php` to apigen command.

Something like:

```
bin/apigen generate --preferred-vendor-autoload other-vendor/autoload.php
```

vlastavesely:
> could your add the parent class to another directory

How do you mean it?
In `repo/src`, I have source files and in `repo/vendor` installed all dependencies. But if I pass `repo/src` to ApiGen, it should ignore all other directories.


Maybe we could assume as default path to autoload path `../vendor` from the directory that is specified for `generate` command...?

But what if I need to generate source from two such repositories at once and with same dependencies...?

TomasVotruba:
> How do you mean it?

I talk about this test:

- create `SomeNamespace\ParentClass` from the test 
- add it to `tests/Generator/NotLoadedSources/ParentClass.php`
- rename it to match PSR-4:  `ApiGen\Tests\Generator\NotLoadedSources\ParentClass`

--

> Maybe we could assume as default path to autoload path ../vendor from the directory that is specified for generate command...?

Yea, that might be much easier :+1: 

That would require only adding `ComposerSourceLocator` with `getcwd() . '/vendor/autoload.php'` as argument. If file would exists.

Could you try that?

vlastavesely:
This is seems to be working.

Yes, I could.

TomasVotruba:
> This is seems to be working.

Thanks, I'll merge it, so we have this covered.

> Yes, I could.

Thanks. If you came up with `.travis.yml` setup which would test it, we could cover that too.

vlastavesely:
You have merget it already? I have pushed the change just now....

vlastavesely:
It worked for the test case we have. But the `ComposerSourceLocator` was not already done.

TomasVotruba:
Oh, I thought you'd create a new branch. I'm sorry.

Could you resend PR?

PR #916, opened 2017-06-30
tests: added class extending from another project

vlastavesely:
next part of #915


packages/Reflection/src/Parser/Parser.php
```diff
@@ -117,10 +118,19 @@ private function parseFunctions(SourceLocator $sourceLocator): void
...
-        return new AggregateSourceLocator([
+        $locators = [
             new DirectoriesSourceLocator($directories),
             new AutoloadSourceLocator(),
             new PhpInternalSourceLocator()
-        ]);
+        ];
+
+        foreach ($directories as $directory) {
+            $autoload = dirname($directory) . '/vendor/autoload.php';
```
vlastavesely: There we should allow to override it by console argument. But I have no idea how :)

TomasVotruba: I think your solution is more elegant. We can use some autodiscovery, like [this](https://github.com/ApiGen/ApiGen/issues/900#issuecomment-311908596).

What do you think?

vlastavesely: Yes, it is more elegant. But what if it does not fit to all use cases...? My one was pretty standard (with `src` and `vendor` in the same directory).

TomasVotruba: You're right this may vary, but I'm happy with current solution.
First I'd add a test for this with a package, that has some extra class that are not in Apigen's vendor:

Something like this a in TestCase. Using [Process](https://github.com/symfony/process) or `exec()`:

```bash
composer create-project nette/application temp/nette-application
bin/apigen generate temp/nette-application
```

Then it would be a small step to extend to fir all use cases.

vlastavesely: Actually, I do not need additional changes to cover all the scenarios possible.

I do not understand with the test :) How do you mean it?

TomasVotruba: Ok :+1: 

I'm writing this from top of my head, but this might be prototype to start with:

```php
final class UseDifferentVendorTest extends AbstactContainerAwareTestCase
{
	protected function setUp()
    {
     	exec('composer create-project nette/application temp/nette-application');

     	$this->parser = $this->container->get(Parser::class);
    }

    public function test()
    {
		$this->parser->parserDirectories([__DIR__ . '/temp/nette-application']);
		// everything works
		$this->assertTrue(true);
    }
}
```

Should fail without your fix :+1:

TomasVotruba:
After this, I think I'll release RC2, so we keep changelog small and easier to test never version.

packages/Reflection/src/Parser/Parser.php
vlastavesely: I will try it.

vlastavesely: That is annoyingly slow.It should be mocked inn some way. I will try to play with it it during a day.

nelson6e65:
I'll wait this to check. In 682fc8f43c463884bfef5fd0c8d56c403c8a5608 for my [php_nml](https://github.com/nelson6e65/php_nml) I'm getting

**`apigen generate src --destination output/api -vvv`**

```
  [ReflectionException (-1)]                          
  Class NelsonMartell\Utilities\Asset does not exist

Exception trace:
 () at /home/ubuntu/.composer/vendor/apigen/apigen/packages/Reflection/src/Reflection/Class_/ClassConstantReflection.php:44
 ReflectionClass->__construct() at /home/ubuntu/.composer/vendor/apigen/apigen/packages/Reflection/src/Reflection/Class_/ClassConstantReflection.php:44
 ApiGen\Reflection\Reflection\Class_\ClassConstantReflection->__construct() at /home/ubuntu/.composer/vendor/apigen/apigen/packages/Reflection/src/Reflection/Class_/ClassConstantReflection.php:56
 ApiGen\Reflection\Reflection\Class_\ClassConstantReflection::createFromNameValueAndClass() at /home/ubuntu/.composer/vendor/apigen/apigen/packages/Reflection/src/Reflection/Class_/ClassReflection.php:226
 ApiGen\Reflection\Reflection\Class_\ClassReflection->getOwnConstants() at /home/ubuntu/.composer/vendor/apigen/apigen/packages/Element/src/Cache/ReflectionWarmUpper.php:33
 ApiGen\Element\Cache\ReflectionWarmUpper->warmUpClassReflections() at /home/ubuntu/.composer/vendor/apigen/apigen/packages/Element/src/Cache/ReflectionWarmUpper.php:24
 ApiGen\Element\Cache\ReflectionWarmUpper->warmUp() at /home/ubuntu/.composer/vendor/apigen/apigen/packages/Reflection/src/Parser/Parser.php:57
 ApiGen\Reflection\Parser\Parser->parseDirectories() at /home/ubuntu/.composer/vendor/apigen/apigen/src/Application/ApiGenApplication.php:55
 ApiGen\Application\ApiGenApplication->runCommand() at /home/ubuntu/.composer/vendor/apigen/apigen/src/Console/Command/GenerateCommand.php:61
 ApiGen\Console\Command\GenerateCommand->execute() at /home/ubuntu/.composer/vendor/symfony/console/Command/Command.php:264
 Symfony\Component\Console\Command\Command->run() at /home/ubuntu/.composer/vendor/symfony/console/Application.php:869
 Symfony\Component\Console\Application->doRunCommand() at /home/ubuntu/.composer/vendor/symfony/console/Application.php:223
 Symfony\Component\Console\Application->doRun() at /home/ubuntu/.composer/vendor/symfony/console/Application.php:130
 Symfony\Component\Console\Application->run() at /home/ubuntu/.composer/vendor/apigen/apigen/bin/apigen:26

generate [--destination DESTINATION] [--] <source> (<source>)...
                                                      
```

**`../ApiGen/bin/apigen generate src --destination output/api -vvv`**

```
PHP Notice:  Undefined offset: 0 in /home/ubuntu/workspace/ApiGen/vendor/symfony/config/Resource/ComposerResource.php on line 68

Notice: Undefined offset: 0 in /home/ubuntu/workspace/ApiGen/vendor/symfony/config/Resource/ComposerResource.php on line 68
PHP Notice:  Undefined offset: 1 in /home/ubuntu/workspace/ApiGen/vendor/symfony/config/Resource/ComposerResource.php on line 68

Notice: Undefined offset: 1 in /home/ubuntu/workspace/ApiGen/vendor/symfony/config/Resource/ComposerResource.php on line 68
                        
  [ReflectionException (-1)]                          
  Class NelsonMartell\Utilities\Asset does not exist  

Exception trace:
// [...]

generate [--destination DESTINATION] [--] <source> (<source>)...
```

vlastavesely:
It was not hard. When I tested this change, I used the generator to generate API for my packages and in both cases it worked in the same way - failed without fix and passed with the change.

But, sadly, console progress is broken again - there is one more line again. What is `EmptyNamespaceGenerator` used for?

TomasVotruba:
I think it's for elements without namespaces.

Like `dump()` function etc.

tests/Generator/ExtendingClassGeneratorTest.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    private $classElementGenerator;
+
+    protected function setUp(): void
+    {
+        /** @var Parser $parser */
+        $parser = $this->container->get(Parser::class);
+        $parser->parseDirectories([__DIR__ . '/ExtendingSources']);
+
+        $this->classElementGenerator = $this->container->get(ClassGenerator::class);
```
TomasVotruba: Here only `ReflectionStorage` is needed for test. Generator can be removed here.

tests/Generator/DifferentVendorSources/vendor/autoload.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+$loader = new Composer\Autoload\ClassLoader();
+
+// Files of the project that API is being generated for
+$loader->add('MyVendor', __DIR__ . '/../src');
+
+// Project's dependencies
+$loader->addClassMap([
```
TomasVotruba: Nice solution!

TomasVotruba:
I think we could merge this, so it's not opened for too long to forget. I spotted only 2 details.

tests/Generator/DifferentVendorSources/vendor/autoload.php
vlastavesely: Thanks :)

tests/Generator/ExtendingClassGeneratorTest.php
TomasVotruba: I'd move it to `package/Reflection/tests/Parser` too.

vlastavesely: :confused: Purpose of the `ExtendingClassGeneratorTest` was to check whether it can process PSR-4 autoloaded classes that are not stored in the `ReflectionStorage`. Should not it be left in `Generator` namespace and assert that html files were really generated? Am I missing something?

TomasVotruba: Well all you actually test is [this line](https://github.com/ApiGen/ApiGen/pull/916/files#diff-9b239be43da25e013c73c057e696bfcaR20), right?

Using *Generator-level test is quite high-level - in testing application it would be Selenium tests.
It's better to use more low-level testing, if possible, like here.

TomasVotruba:
Merging :hammer:  , thank you for you work

PR #919, opened 2017-07-01
Theme: split classes and exceptions in namespace summary

vlastavesely:
In older ApiGen, there were classes and exceptions summaries separated.

## Before:
before

## After
after

### DIFF
--- a/packages/Reflection/src/ReflectionStorage.php
+++ b/packages/Reflection/src/ReflectionStorage.php
@@ -6,6 +6,7 @@ use ApiGen\Reflection\Contract\Reflection\Class_\ClassReflectionInterface;
 use ApiGen\Reflection\Contract\Reflection\Function_\FunctionReflectionInterface;
 use ApiGen\Reflection\Contract\Reflection\Interface_\InterfaceReflectionInterface;
 use ApiGen\Reflection\Contract\Reflection\Trait_\TraitReflectionInterface;
+use Throwable;
 
 final class ReflectionStorage
 {
@@ -13,6 +14,11 @@ final class ReflectionStorage
      * @var ClassReflectionInterface[]
      */
     private $classReflections = [];
+
+    /**
+     * @var ClassReflectionInterface[]
+     */
+    private $exceptionReflections = [];
 
     /**
      * @var InterfaceReflectionInterface[]
@@ -38,6 +44,14 @@ final class ReflectionStorage
     }
 
     /**
+     * @return ClassReflectionInterface[]
+     */
+    public function getExceptionReflections(): array
+    {
+        return $this->exceptionReflections;
+    }
+
+    /**
      * @param ClassReflectionInterface[] $classReflections
      */
     public function addClassReflections(array $classReflections): void
@@ -46,7 +60,11 @@ final class ReflectionStorage
         });
         sort($classReflections);
         foreach ($classReflections as $classReflection) {
-            $this->classReflections[$classReflection->getName()] = $classReflection;
+            if ($classReflection->implementsInterface(Throwable::class)) {
+                $this->exceptionReflections[$classReflection->getName()] = $classReflection;
+            } else {
+                $this->classReflections[$classReflection->getName()] = $classReflection;
+            }
         }
     }
 
--- a/packages/ThemeDefault/src/@elementlist.latte
+++ b/packages/ThemeDefault/src/@elementlist.latte
@@ -36,3 +36,10 @@
         {include elements, elements => $allFunctions}
     </table>
 {/if}
+
+{if $exceptions}
+    <table class="summary table table-responsive table-bordered table-striped" id="exceptions">
+        <tr><th colspan="2">Exceptions Summary</th></tr>
+        {include elements, elements => $exceptions}
+    </table>
+{/if}
--- a/packages/ThemeDefault/src/partial/menu.latte
+++ b/packages/ThemeDefault/src/partial/menu.latte
@@ -34,6 +34,12 @@
                             </li>
                         {/if}
 
+                        {if count($allExceptions)}
+                            <li>
+                                <a n:class="$activePage === exceptions ? active" href="exceptions.html">Exceptions</a>
+                            </li>
+                        {/if}
+
                         {foreach $annotationGroups as $annotationGroup}
                             <li class="separator"></li>
                             <li>
--- a/src/Console/Progress/StepCounter.php
+++ b/src/Console/Progress/StepCounter.php
@@ -30,6 +30,7 @@ final class StepCounter
         return $this->getSourceCodeStepCount()
             + count($this->namespaceReflectionCollector->getNamespaces())
             + count($this->reflectionStorage->getClassReflections())
+            + count($this->reflectionStorage->getExceptionReflections())
             + count($this->reflectionStorage->getTraitReflections())
             + count($this->reflectionStorage->getInterfaceReflections())
             + count($this->reflectionStorage->getFunctionReflections())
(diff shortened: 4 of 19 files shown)


src/Generator/NamespaceGenerator.php
```diff
@@ -45,6 +45,16 @@ private function generateForNamespace(
         string $namespace,
         NamespaceReflectionCollector $namespaceReflectionCollector
     ): void {
+        $classes = $exceptions = [];
+
+        foreach ($namespaceReflectionCollector->getClassReflections($namespace) as $class) {
+            if (strtolower(substr($class->getName(), -9)) !== 'exception') {
```
TomasVotruba: Here can be used `is_a` function to check if the class is exception:

```php
is_a($class, Exception::class, true)
```

Name is less reliable than this.

TomasVotruba:
I like this!

Just one detail and it's ready

packages/ThemeDefault/src/@elementlist.latte
```diff
@@ -36,3 +36,10 @@
         {include elements, elements => $allFunctions}
     </table>
 {/if}
+
+{if $exceptions}
```
TomasVotruba: Are `EmptyNamespaceGenerator` and `NamespaceGenerator` only places where this template `@elementlist.latte` is used?

I think it's also in [IndexGenerator template](https://github.com/ApiGen/ApiGen/blob/682fc8f43c463884bfef5fd0c8d56c403c8a5608/packages/ThemeDefault/src/index.latte#L20)

Maybe it would make sense to distinquish those already on `ReflectionStorage` level: https://github.com/ApiGen/ApiGen/blob/682fc8f43c463884bfef5fd0c8d56c403c8a5608/packages/Reflection/src/ReflectionStorage.php#L81

vlastavesely:
Oh yes, moving into `ReflectionStorage` makes sense. I'll do it.

vlastavesely:
What do you prefer on `classes.html` page - list of classes and list of exceptions under it or new page `exceptions.html`?

src/Generator/NamespaceGenerator.php
vlastavesely: I tried it but it does not seem to work. What about `$classReflection->implementsInterface('Throwable')`? That covers not only classes derived from `Exception` but also any other throwable classes that may not inherit from the `Exception`.

TomasVotruba:
Standalone exception page would make more sense, as it would be 1:1 to `ReflectionStorage`.

vlastavesely:
* added exceptions summary in `exceptions.html`
* link to the summary added into header (after "Overview", "Classes"...)
* added `ExceptionGenerator` - it uses `class.latte` (for exception is class actually)
* `exceptions.html` added into StepCountter but still, there is 1 item missing in the counter :(
* `ExceptionsGenerator` and `ExceptionGenerator` covered by tests

packages/Reflection/src/ReflectionStorage.php
```diff
@@ -46,7 +59,11 @@ public function addClassReflections(array $classReflections): void
         });
         sort($classReflections);
         foreach ($classReflections as $classReflection) {
-            $this->classReflections[$classReflection->getName()] = $classReflection;
+            if ($classReflection->implementsInterface('Throwable')) {
```
TomasVotruba: There could be used class instead of string: `Throwable::class`, it's easier to refactor and more future proof

packages/ThemeDefault/src/partial/menu.latte
```diff
@@ -34,6 +34,12 @@
                             </li>
                         {/if}
 
+                        {if count($allExceptions)}
+                            <li>
+                                <a n:class="$activePage === exceptions ? active" href="exceptions.html">Exceptions</a>
+                            </li>
+                        {/if}
```
TomasVotruba: Very nice!

src/EventSubscriber/ElementsTemplateVariablesEventSubscriber.php
```diff
@@ -53,6 +53,7 @@ public function loadTemplateVariables(CreateTemplateEvent $createTemplateEvent):
         $parameterBag->addParameters([
             'allNamespaces' => $this->namespaceReflectionCollector->getNamespaces(),
             'allClasses' => $this->reflectionStorage->getClassReflections(),
+            'allExceptions' => $this->reflectionStorage->getExceptionReflections(),
```
TomasVotruba: That's the right place, I'm glad it's obvious.
:+1:

src/Console/Progress/StepCounter.php
```diff
@@ -51,6 +53,10 @@ private function getOverviewPagesCount(): int
             $count++; // classes.html
         }
 
+        if (count($this->reflectionStorage->getExceptionReflections())) {
+            $count++; // exceptions.html
```
TomasVotruba: :+1:  for comment

src/Generator/ExceptionGenerator.php
```diff
@@ -0,0 +1,82 @@
...
+            $this->generateForException($exceptionReflection);
+            $this->generateSourceCodeForException($exceptionReflection);
+        }
+    }
+
+    private function generateForException(ClassReflectionInterface $exceptionReflection): void
+    {
+        $this->templateRenderer->renderToFile(
+            $this->configuration->getTemplateByName('class'),
+            $this->configuration->getDestinationWithPrefixName('class-', $exceptionReflection->getName()),
```
TomasVotruba: I would probably prefer `exception-` here, to be 1:1 with ReflectionStorage.

It was and might be confusing in the future, if something is class x exception and something is only class.

tests/Generator/ExceptionGeneratorTest.php
```diff
@@ -0,0 +1,36 @@
...
+        $this->exceptionElementGenerator->generate();
+
+        $this->assertFileExists(
+            TEMP_DIR . '/class-ApiGen.Tests.Generator.Source.SomeException.html'
+        );
+        $this->assertFileExists(
+            TEMP_DIR . '/source-class-ApiGen.Tests.Generator.Source.SomeException.html'
+        );
+    }
+}
```
TomasVotruba: Beautiful test :clap:

TomasVotruba:
> exceptions.html added into StepCountter but still, there is 1 item missing in the counter :(

What exactly do you mean by that? I'd like to help you solve it.

packages/Reflection/src/ReflectionStorage.php
vlastavesely: ok

vlastavesely:
I mean that console output looks like this:
```
 21/21 [============================] 100% 2 secs/2 secs 8.0 MiB - Finished!
 22/22 [============================] 100% 2 secs/2 secs 8.0 MiB - Finished!
```
Works well but on end it prints one more line. We already know where to fix it (see #905) but what file generated is doing this....? I do not know.

src/Generator/ExceptionGenerator.php
vlastavesely: ok, working on it.

vlastavesely:
* `class-` prefix changed to `exception-`
* fixed StringRouter
* covered by tests

TomasVotruba:
Thank you for changes, I'll look at progress bar personally.

src/Generator/ExceptionGenerator.php
TomasVotruba: Thank you

vlastavesely:
fixed cs

TomasVotruba:
:clap:  

Thank you, merged!

vlastavesely:
Have you checked the progressbar issue? It seems to be namespace-related. I just added a class from another namespace (`Nette\Loaders`) and there appeared one new line so now they are three and not just two like yesterday.
```
 26/26 [============================] 100% 8 secs/8 secs 8.0 MiB - Finished!
 27/27 [============================] 100% 8 secs/8 secs 8.0 MiB - Finished!
 28/28 [============================] 100% 8 secs/8 secs 8.0 MiB - Finished!
```
Are the sub-namespaces counted?

TomasVotruba:
Not yet, I'm busy with my work now so I don't have time.

I still think there will be some simple condition issue.

vlastavesely:
Found it! Tomorrow, I will fix it. When namespace is empty (does not contain classes), it is not counted.

PR #921, opened 2017-07-01
Improvements in binary autoloading

nelson6e65:
- [x] Move auto-loading functionality to `bin/bootstrap.php`.
- [x] Throws a `RuntimeException` exception if ApiGen was not able to find its autoloader.
- [x] Include `bin/` directory to be checked for CS issues from `composer complete-check`.


bootstrap/autoloader.php
```diff
@@ -0,0 +1,29 @@
...
+
+if ($local_autoloader) {
+    $cli_autoloader = $local_autoloader;
+    define('APIGEN_CLI_CONTEXT', APIGEN_CLI_LOCAL_CONTEXT);
+} elseif ($dist_autoloader) {
+    $cli_autoloader = $dist_autoloader;
+    define('APIGEN_CLI_CONTEXT', APIGEN_CLI_DIST_CONTEXT);
+} else {
+    $cli_autoloader = null;
+    define('APIGEN_CLI_CONTEXT', null);
```
TomasVotruba: Constants that are used locally have no added value here.
Could you remove them please?

bootstrap/autoloader.php
```diff
@@ -0,0 +1,29 @@
...
+
+    throw new RuntimeException('Error! Unable to find ApiGen autoloader.');
+}
+
+require_once $cli_autoloader;
```
TomasVotruba: Please use $camelCase. 

This file should be checked by coding standard as well. Could you add it to `composer.json`?

bin/apigen
```diff
@@ -7,17 +7,7 @@ use Symfony\Component\Console\Application;
...
-    __DIR__ . '/../../../autoload.php'
-];
-
-foreach ($possibleAutoloadPaths as $possibleAutoloadPath) {
-    if (file_exists($possibleAutoloadPath)) {
-        require_once $possibleAutoloadPath;
-        break;
-    }
-}
+require_once __DIR__.'/../bootstrap/autoloader.php';
```
TomasVotruba: It's common practise to pu this `autoloader.php` to `bin/bootstrap.php` for console application.
No need to create extra directory that is used only here.

bootstrap/autoloader.php
```diff
@@ -0,0 +1,29 @@
...
+define('APIGEN_CLI_ROOT_DIR', realpath(__DIR__.'/../'));
+
+// Cloned, downloaded or installed via `composer create-project`
+define('APIGEN_CLI_LOCAL_CONTEXT', 'local');
+// Installed via `composer [global] require`
+define('APIGEN_CLI_DIST_CONTEXT', 'distributed'); // Installed via composer require
+
+
+$local_autoloader = realpath(APIGEN_CLI_ROOT_DIR.'/vendor/autoload.php');
+$dist_autoloader  = realpath(APIGEN_CLI_ROOT_DIR.'/../../autoload.php');
```
TomasVotruba: I'd stick with simple definitions, I get lost in these. There is no need for `realpath()` here. Maybe add a comment behind those, to make it clear:

```php
$possibleAutoloadPaths = [
    __DIR__ . '/../vendor/autoload.php', // local
    __DIR__ . '/../../../autoload.php' // dist
];
``` 

What dost local/dist actually mean? I see them as dependency - `composer require` and project `composer create-project`. Is there any other difference?

composer.json
```diff
@@ -67,8 +67,8 @@
     "scripts": {
         "complete-check": ["phpunit", "@phpstan", "@check-cs"],
         "phpstan": "phpstan analyse packages src --level 4 -c phpstan.neon",
-        "check-cs": "easy-coding-standard check packages src",
-        "fix-cs": "easy-coding-standard check packages src --fix"
+        "check-cs": "easy-coding-standard check packages src bin bootstrap",
+        "fix-cs": "easy-coding-standard check packages src bin bootstrap --fix"
```
TomasVotruba: While you're at it: could you add tests here? They're skipped now, I probably forgot to add them.

bootstrap/autoloader.php
```diff
@@ -0,0 +1,29 @@
...
+    $cli_autoloader = $local_autoloader;
+    define('APIGEN_CLI_CONTEXT', APIGEN_CLI_LOCAL_CONTEXT);
+} elseif ($dist_autoloader) {
+    $cli_autoloader = $dist_autoloader;
+    define('APIGEN_CLI_CONTEXT', APIGEN_CLI_DIST_CONTEXT);
+} else {
+    $cli_autoloader = null;
+    define('APIGEN_CLI_CONTEXT', null);
+
+    throw new RuntimeException('Error! Unable to find ApiGen autoloader.');
```
TomasVotruba: This exception message could be more informative how to solve it, e.g. "run composer update"

TomasVotruba:
I like separation of concerns to another file :+1:

bootstrap/autoloader.php
nelson6e65: I defined here to for, for example, if you catch the exception, you can still get defined the constant with `null`. Not now, but... maybe later. :grey_question:

Variables, yes. Maybe not useful.

bin/apigen
nelson6e65: Nice! I'll do.

bootstrap/autoloader.php
nelson6e65: Yes. I was thinking about.. dome checks later with this constants. But maybe is more useful a boolean value for the constant.

I made some checks and only using `composer [global] require` the `$dist_autoloader` is used. So, in future, you can use this info.


About the `realpath()`... This actually check if the file is accesible and set to `false` if was not found or there is not permissions. Besides the path is more clear, absolute and consistent with OS `/` and `\`  directory separator (no need concatenate many `DIRECTORY_SEPARATOR`).

bootstrap/autoloader.php
TomasVotruba: Using global constants usually means bad design that leads to static hell. 

I've learned that hard way using [CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) that uses constant in various parts.

They're very difficult to change once set up. It's difficult to test them - to set various values for them.

nelson6e65: Maybe using `putenv()` instead of global constants, [as composer does](https://github.com/composer/composer/blob/master/bin/composer#L51)?

TomasVotruba:
Thank you for changes.

What issue do you try to solve by using constants?

nelson6e65:
I don't know, but it should be useful to know the real current path of binary or root directory. For example, you could simplify this lines of [`ThemeDirectoryOption`](https://github.com/ApiGen/ApiGen/blob/master/packages/ModularConfiguration/src/Option/ThemeDirectoryOption.php#L43-L54) to 

```php
return APIGEN_CLI_ROOT_DIR.'/packages/ThemeDefault/src';
```

In fact, I will check this right now. :grinning:

TomasVotruba:
I don't want to promote using global constans as I see it bad practise as [I described above](https://github.com/ApiGen/ApiGen/pull/921#discussion_r125169950).

As you can see, it can easily spread everywhere in the code. Not a way to go. 

If you need this, it should be set by parameter - using `ParameterProvider`. Same as `$appDir` in Nette or `$kernelRootDir` in Symfony.

nelson6e65:
I see. I'll remove then, since there is no need for constants for this file only. :alien:

composer.json
```diff
@@ -66,9 +66,9 @@
     },
     "scripts": {
         "complete-check": ["phpunit", "@phpstan", "@check-cs"],
-        "phpstan": "phpstan analyse packages src --level 4 -c phpstan.neon",
-        "check-cs": "easy-coding-standard check packages src",
-        "fix-cs": "easy-coding-standard check packages src --fix"
+        "phpstan": "phpstan analyse packages src bin --level 4 -c phpstan.neon",
+        "check-cs": "easy-coding-standard check packages src bin",
```
TomasVotruba: Could add `tests` to check here as well? I probably forgot to add it:

```bash
check bin packages src tests
```

bin/bootstrap.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
```
TomasVotruba: Could you put this on 1st line as in other files? I realize here coding standard check doesn't work.

bin/bootstrap.php
```diff
@@ -0,0 +1,22 @@
...
+$autoloader = realpath($rootDir.'/vendor/autoload.php');
+
+if (!$autoloader) {
+    // Installed via `composer [global] require`.
+    $autoloader = realpath($rootDir.'/../../autoload.php');
+}
+
+if (!$autoloader) {
+    throw new RuntimeException(
+        'Error! ApiGen was unable to find its autoloader. '.
```
TomasVotruba: Just a detail: "Error! " can be ommited, it's obvious from Exception that something is wrong. No need to say it twice.
It would be more pleasant to read without it.

bin/bootstrap.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+$rootDir = realpath(__DIR__.'/../');
+
+// Autoloader in a standalone install (not installed via `composer [global] require`)
```
TomasVotruba: As there are more options to install package, it would make more sense **what it IS** than what it isn't. E.g.:

```
installed via `composer require`
```

TomasVotruba:
Thank you! 

Just a few details.

composer.json
nelson6e65: I'll add with CS issues fixed.

nelson6e65:
I think adding `tests` for CS checks should be made in another PR, since there are many errors.  :sweat_smile:
https://github.com/ApiGen/ApiGen/pull/921#discussion_r125181528

composer.json
nelson6e65: @TomasVotruba Well... I think _adding `tests` for CS checks_ should be made in another PR after this get merged, since there are many errors in `tests/` that are not related with this PR to be solved.  :sweat_smile:

https://github.com/ApiGen/ApiGen/pull/921#discussion_r125181528

For example, CS should ignore files in `tests/Generator/Sources/`.

nelson6e65: If I add tests in there, checks will not pass for this PR.

TomasVotruba: Ok, you're right. New PR is better idea :+1:

TomasVotruba:
Thank you for great work and fast responds to my comment :cl

PR #922, opened 2017-07-02
Improve done and fail messages and workaround for breaking progressbar

nelson6e65:
- [x] Add workaround for maximum steps counter progress :bug: which break the progress-bar.
- [x] Add :new: message for failing state.
- [x] Improve the "finished" message.
- [x] :art: Add colours for success and fail messages.

### DIFF
--- a/src/Console/Helper/ProgressBar.php
+++ b/src/Console/Helper/ProgressBar.php
@@ -41,8 +41,18 @@ final class ProgressBar
         }
 
         $this->bar->advance($increment);
-        if ($this->bar->getProgress() === $this->bar->getMaxSteps()) {
-            $this->output->writeln(' - Finished!');
+    }
+
+    public function __destruct()
+    {
+        if ($this->bar) {
+            if ($this->bar->getProgress() === $this->bar->getMaxSteps()) {
+                $msg = '<info>Done! API generation was successfully completed.</info> ';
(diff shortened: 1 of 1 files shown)


src/Console/Helper/ProgressBar.php
```diff
@@ -41,8 +41,18 @@ public function increment(int $increment = 1): void
         }
 
         $this->bar->advance($increment);
-        if ($this->bar->getProgress() === $this->bar->getMaxSteps()) {
```
TomasVotruba: I think better solution would be in this condition to stop it of overload. I'd try something like:

```php
if ($this->bar->getProgress() >= $this->bar->getMaxSteps()) {
}
```

nelson6e65: The #902 had this 5 "extra steps". This is not a problem about the maximum is set in the [`init()`](https://github.com/ApiGen/ApiGen/blob/ad95bd33bb2869bed7f19061ab88d092389926a1/src/Console/Helper/ProgressBar.php#L30)?

#905

vlastavesely:
After #927 gets merged, progressbar issue should be fixed for ever and ever. I hope :)

TomasVotruba:
Replaced by #927 

Thank you for your work anyway!

PR #923, opened 2017-07-02
Theme: used local Bootstrap CSS

vlastavesely:
From time to time, I am working offline. Using remote CSS is a pain when no connection is available or when I would like to avoid mobile data depleting. :)

Also, it should be quicker. With the remote version it was changing the layout when remote CSS was downloaded and parsed.

The CSS I removed was not used. It looks like some old version that was so different from the `bootstrap.min.css`.


packages/ThemeDefault/src/@layout.latte
```diff
@@ -6,8 +6,7 @@
         <meta name="mobile-web-app-capable" content="yes">
         <meta name="apple-mobile-web-app-capable" content="yes">
 
-        {* Bootstrap 3 CDN *}
-        <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous">
+        <link rel="stylesheet" href="resources/bootstrap.min.css">
```
TomasVotruba: Could you mention version and link here in comments? Just to store the version.

vlastavesely: Added. The version number is also inside CSS itself.

TomasVotruba: Thanks! 
I'm aware of that, it's just more clear and accessible right above the file inclusion.

TomasVotruba:
Thank you!

PR #925, opened 2017-07-02
Theme: fixed whitespace wrapping for long description

vlastavesely:
## Before
before

## After
after

### DIFF
--- a/packages/ThemeDefault/src/partial/method.latte
+++ b/packages/ThemeDefault/src/partial/method.latte
@@ -4,7 +4,7 @@
 <tr id="_{$method->getName()}" n:class="$method->isDeprecated() ? deprecated">
     {var $annotations = $method->getAnnotations()}
 
-    <td class="attributes">
+    <td class="attributes col-md-1">
         <code>
             {if $isInterface === true}
                 {ifset $annotations['return']}{$annotations['return'][0]|annotation:$method|noescape}{/ifset}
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thanks for the fix. Have you tested that on mobile? It think it might be related.

It would also make sense to make "public" column much shorted. (Bad idea: maybe using `col-md-1`?)

vlastavesely:
On mobile? No,I just resized browser window and after my fix it looks good.

But I've found another problem :) In the older ApiGen, there was return typehint in the first column:

hint

It needs to be fixed also :) I will do that tomorrow and then we can shorten the first column if needed.

TomasVotruba:
Allright :). Thanks for your weekend time on this. Much appreciated :+1:

vlastavesely:
Could you merge it as it is - without the return types? For new changes I need the changes we already merged after I started working on this. I would start with new PR.

After the changes from this PR, it looks like this:
screenshot-1

TomasVotruba:
Right align was removed for bad readability in case of extra values, see: https://github.com/ApiGen/ApiGen/pull/925#issuecomment-312515770

I'd suggest going for `<td class="col-md-1">` for 1st column and `<nobr>` for it's content. It might look very nice without any extra css.

Could you try that?

vlastavesely:
Two examples of current state. First one is on small screen, second one on a wide screen.
screenshot-1

TomasVotruba:
Looks good to me. Can i Merge it?

vlastavesely:
yes.

TomasVotruba:
Done! Thansk

PR #927, opened 2017-07-03
fixed console progressbar

vlastavesely:
I hope this is fixed for good. It was caused by empty parent namespaces - at least I know what the `EmptyNamespaceGenerator` is for :laughing:


TomasVotruba:
Thanks for the PR. We don't need to hope, we could add a test :+1: 

What do you think?

vlastavesely:
I think it is really good idea with the test! :laughing:

TomasVotruba:
Let's go for it then :+1: 

Thank you

Issue #881, opened 2017-06-10
[5.x][help needed] Fix autocomplete javascript


TomasVotruba:
There was support for class, interface, trait, function and namespace autocomplete in search input.

## How it Should Work

1. Start typing a class name (fuzzy search will happen)
2. Pick your class by arrows or mouse
3. Hit "enter"
3. It will redirect to url

You can try it yourself in [Doctrne API](http://www.doctrine-project.org/api/orm/2.5/) or just watch me recoding it for you:

![aae](https://user-images.githubusercontent.com/924196/27001982-dc04c93e-4dd6-11e7-8c6e-5c67bc137d35.gif)

I've upgrade to new Jquery 3.1, changed format of original data provided by `elementlist.js` file and dropped lof of Javascript - this all broke it. I'm no Javascript master, so need help with this.

## Current `elementlist.js` format 

- element name => filepath to redirect

```javascript
var ApiGen = ApiGen || {};
ApiGen.elements = {
    "ApiGen\\Tests\\Generator\\Source":"namespace-ApiGen.Tests.Generator.Source.html",
    "ApiGen\\Tests\\Generator\\Source\\someFunction()":"function-ApiGen.Tests.Generator.Source.someFunction.html",
    "ApiGen\\Tests\\Generator\\Source\\SomeClass":"class-ApiGen.Tests.Generator.Source.SomeClass.html",
    "ApiGen\\Tests\\Generator\\Source\\SomeInterface":"interface-ApiGen.Tests.Generator.Source.SomeInterface.html",
    "ApiGen\\Tests\\Generator\\Source\\SomeTrait":"trait-ApiGen.Tests.Generator.Source.SomeTrait.html"   
};
```

## Important files with autocomplete source code

- [elementlist.js](https://github.com/ApiGen/ApiGen/blob/master/packages/ThemeDefault/src/elementlist.js.latte) - this is template for generating data in format mentioned above
- [resources/main.js](https://github.com/ApiGen/ApiGen/blob/8a7e1399532fdf0709910f6888fe6944f9054153/packages/ThemeDefault/src/resources/main.js#L8-L72) - javascript code for old autocomplete, I expect it to be much simpler now, as before URL was generated by javascript parsing and now it's the name in array.
- jsfiddle for Jquery 3.1 and JqueryUI with autocomplete - https://jsfiddle.net/jphellemons/0ukbtgs4/
[...]

PR #931, opened 2017-07-06
Implemented fuzzy search

POPSuL:
Related issue: https://github.com/ApiGen/ApiGen/issues/881
Examples:
image
image
image


TomasVotruba:
Thanks for submitted PR.

Does it support highlighting of searched text?

Can I pick up solution by mouse and keyboard?

TomasVotruba:
See #932 for other solution.

I like it's simplicity and text-highlighting, even if doesn't fuzzy search items.

POPSuL:
@TomasVotruba highlighting is not implemented yet.
Items can be picked both keyboard (type some text -> down/up -> enter) and mouse (just click)

POPSuL:
@TomasVotruba I don't mind this (#932) decision if it implements a fuzzy search :)

TomasVotruba:
Ok, let's continue there then :+1: 
We can take it gradually

Thanks for your openness

PR #932, opened 2017-07-06
added autocomplete search

vlastavesely:
Features:
* lightweight as possible (in fact this is a few lines of code) and therefore easy to understand
* highlighting matches
* no levenshtein (filtering is done by jQuery)

screenshot-1

Tested in FF and Chromium


vlastavesely:
It breaks tests (changed structure of `elementlist.js`) so it must be fixed (if this PR is going to be used, otherwise it is useless...).

TomasVotruba:
I must say I like it. Code is quite reable even for me and I don't use Javascript at all.

Can I use keyboard to get to pick the element from autocomplete or is mouse needed?

vlastavesely:
No matter, both can be used - keyboard or mouse.

TomasVotruba:
That's great. 

Is there support for fuzzy search? 
When I type "em" => "EntityManager" will be found as well.

I'm not sure it is needed so much to worth the complexity, but anaway.

vlastavesely:
No, I think that the default behavior of the jQuery autocomplete is not 
able to do it.

TomasVotruba:
> But it is really needed? Until now, I was not aware that the original search worked that way :)

I'm not sure I ever used that too :)

I was curious about concurrency, so I've check [phpdocumentor demo](https://demo.phpdoc.org/Clean/namespaces/phpDocumentor.html) - no search, [sami demo](http://api.symfony.com/3.3/index.html) - normal search.

But if it would be easy to implement, I'd go for it... Could give it 15 min tops? If that won't be enought, we don't need it.

packages/ThemeDefault/src/resources/main.js
```diff
@@ -2,75 +2,23 @@
...
-    // Content
-/*
-    var availableTags = [
-        "ActionScript",
-        "Able"
-    ];
+    // Autocomplete search
+    $.ui.autocomplete.prototype._renderItem = function (ul, item) {
+        var re = new RegExp('(' + this.term.replace(/\\/g, '\\\\') + ')', 'i');
+        var t = item.label.replace(re, "<b>$1</b>");
```
TomasVotruba: Could you please expand names like `re` and `t`, to be more expressive and readable. 

I recall this was pain to debug for me, because of ambiguous naming.

packages/Element/src/AutocompleteElements.php
```diff
@@ -64,19 +64,19 @@ private function addReflections(array $elements): array
...
         }
 
         foreach ($this->reflectionStorage->getInterfaceReflections() as $interfaceReflection) {
-            $elements[$interfaceReflection->getName()] = $this->reflectionRoute->constructUrl($interfaceReflection);
+            $elements[] = ['file' => $this->reflectionRoute->constructUrl($interfaceReflection), 'label' => $interfaceReflection->getName()];
         }
 
         foreach ($this->reflectionStorage->getTraitReflections() as $traitReflection) {
-            $elements[$traitReflection->getName()] = $this->reflectionRoute->constructUrl($traitReflection);
+            $elements[] = ['file' => $this->reflectionRoute->constructUrl($traitReflection), 'label' => $traitReflection->getName()];
```
TomasVotruba: Could you put this array on multiline? It's more readable.

```php
item => value
item2 => value
```

I wish there [would be a fixer for that](https://github.com/Symplify/Symplify/issues/59#issuecomment-313641755).

vlastavesely:
OK, let us try it.

packages/ThemeDefault/src/resources/main.js
```diff
@@ -2,75 +2,23 @@
...
+            .appendTo(ul);
+    };
 
-    $('#search input[name=q]').autocomplete({
-        source: availableTags // ApiGen.elements
+    $("#search input[name=q]").autocomplete({
+        source: ApiGen.elements,
+        select: function (event, ui) {
+            window.location.href = ui.item.file;
+        }
```
TomasVotruba: I wish all javascript I read was that simple :clap: 

Love it

packages/ThemeDefault/src/resources/main.js
vlastavesely: `t` - ok, but `re`? I think it is pretty much normal to name regular expression `re` :D an tips for better name?

TomasVotruba: I see. 

`regularExpression` would be more helpful, I didn't event continue to read to `RegExp`, because of the naming :)

Shortcuts are quite nasty limits in communication, because one expects that everyone know them. The problem is, everybody has his own idea and knowledge of those shortcuts. I feel bit stupid now not knowing your shortcut :D 

Regualar Expresion for what exactly? 

Compare:

```php
const RE = '/**()+.[].(d+)w_..*/; # hm, mess?
const PATTERN = '/**()+.[].(d+)w_..*/; # better, probably somehting to match
const PATTERN_SEARCH_IN_ELEMENTS = '/**()+.[].(d+)w_..*/; # the best of those
```

`regularExpression` would be better. More specific the best :+1:

vlastavesely: `highlightPattern` :heart:

vlastavesely:
Now, I will try the find out how hard would be to implement the fuzzy search.

vlastavesely:
To implement the fuzzy search seems not to be impossible since the filter function can be replaced. But I am sure it would not be 15 minutes :D

How about to leave it with the normal search and the fuzzy thing try in the future...?

packages/Element/tests/AutocompleteElementsTest.php
```diff
@@ -41,30 +41,68 @@ protected function setUp(): void
...
 
         $classReflections = $this->reflectionStorage->getClassReflections();
         $this->assertCount(2, $classReflections);
 
         foreach ($classReflections as $classReflection) {
-            $this->assertContains($this->reflectionRoute->constructUrl($classReflection), $autocompleteElements);
+            $this->assertTrue($this->autocompleteHasItem(
+                $this->reflectionRoute->constructUrl($classReflection),
+                $autocompleteElements, 'file'
+            ));
```
TomasVotruba: What exactly is this change for?

Maybe there would be more useful `@dataProvider`, it would keep code more clear.

[You can inspire here from other test](https://github.com/ApiGen/ApiGen/blob/6d8fd611f24c166ddd2e8dea28efc3d46b2a7466/packages/StringRouting/tests/Route/ReflectionRouteTest.php#L45-L70).

packages/Element/tests/AutocompleteElementsTest.php
```diff
@@ -41,30 +41,68 @@ protected function setUp(): void
         $this->reflectionRoute = $this->container->get(ReflectionRoute::class);
     }
 
+    /**
+     * @param string[][]  $items
+     */
+    private function autocompleteHasItem(string $pattern, array $items, string $what = 'label'): bool
```
TomasVotruba: "getters"/"hasser" should start with the verb: `hasAutocompleteItem()`
That way it's more clear that return value is bool. `autocomplete<>` suggest some operation over passed value.

```
isReady()
hasName()
```

TomasVotruba: Also, I'd ommit optional parameter `string $what = 'label'`.
It's form of hidden dependency and I'm not sure what is that for even from reading a code.

packages/Element/tests/AutocompleteElementsTest.php
```diff
@@ -41,30 +41,68 @@ protected function setUp(): void
...
+    }
+
     public function testGetElementsClasses(): void
     {
         $autocompleteElements = $this->autocompleteElements->getElements();
         $this->assertCount(8, $autocompleteElements);
 
-        $this->assertArrayHasKey($this->namespacePrefix, $autocompleteElements);
-        $this->assertArrayHasKey($this->namespacePrefix . '\SubNamespace', $autocompleteElements);
+        $this->assertTrue($this->autocompleteHasItem(
```
TomasVotruba: **Oh, I might get it! :)** Maybe you're looking for `$this->assertContains()`.

```php
$this->assertContains(this->namespacePrefix, $autocompleteElements);
```

If that doesn't work, let me know what exactly is the goal here. I think there will be an assert for it somewhere :sun_behind_small_cloud:

vlastavesely: the item is array and not a string (like `$this->namespacePrefix`) so it should not pass with the `assertContains`.

TomasVotruba:
Alright, let's take it step by step 

Thanks for the fast fix-ups

PR #933, opened 2017-07-06
used local copy of jQuery

vlastavesely:
The same thing as with the Bootstrap theme. When API is browsed offline it would not work well without this.


TomasVotruba:
I see. Thanks for version links!

PR #934, opened 2017-07-06
disabled function param sorting + fixed function.latte

vlastavesely:
Method
```php
  public function test(Test $self, bool $bool = TRUE)
  {
  }
```
cannot be rendered as:
bad

This does not even make any sense for the function cannot be invoked with `bool` as first parameter. Instead, API reference must respect the parameters order:

good

### DIFF
--- a/packages/Reflection/src/TransformerCollector.php
+++ b/packages/Reflection/src/TransformerCollector.php
@@ -64,7 +64,6 @@ final class TransformerCollector
             $elements[$transformedReflection->getName()] = $transformedReflection;
         }
 
-        // @todo: sort here!, before ElementSorter
         $matchingTransformer = $this->detectMatchingTransformer($reflections);
 
         if ($matchingTransformer instanceof SortableTransformerInterface) {
--- a/packages/Reflection/src/TransformerCollector.php
+++ b/packages/Reflection/src/TransformerCollector.php
@@ -35,7 +35,7 @@ final class TransformerCollector
      * @param object[] $reflections
      * @return object[]
      */
-    public function transformGroup(array $reflections, bool $sort = true): array
+    public function transformGroup(array $reflections): array
     {
         $elements = [];
         foreach ($reflections as $name => $reflection) {
@@ -48,11 +48,11 @@ final class TransformerCollector
         }
 
         // @todo: sort here!, before ElementSorter
-        if ($sort) {
+        //if ($sort) { FIXME
             uasort($elements, function ($firstElement, $secondElement) {
                return strcmp($firstElement->getName(), $secondElement->getName());
             });
-        }
+        //}
 
         return $elements;
     }
--- a/packages/Reflection/src/TransformerCollector.php
+++ b/packages/Reflection/src/TransformerCollector.php
@@ -5,6 +5,7 @@ namespace ApiGen\Reflection;
 use ApiGen\Element\ReflectionCollectorCollector;
 use ApiGen\Reflection\Contract\Reflection\Partial\AccessLevelInterface;
 use ApiGen\Reflection\Contract\Reflection\Partial\AnnotationsInterface;
+use ApiGen\Reflection\Contract\Transformer\SortableTransformerInterface;
 use ApiGen\Reflection\Contract\Transformer\TransformerInterface;
 use ApiGen\Reflection\Contract\TransformerCollectorAwareInterface;
 use ApiGen\Reflection\Exception\UnsupportedReflectionClassException;
@@ -31,6 +32,19 @@ final class TransformerCollector
         $this->transformers[] = $transformer;
     }
 
+    private function detectMatchingTransformer(array $reflections): ?TransformerInterface
+    {
+        $reflection = array_shift($reflections);
+
+        foreach ($this->transformers as $transformer) {
+            if ($transformer->matches($reflection)) {
+                return $transformer;
+            }
+        }
+
+        return null;
+    }
+
     /**
      * @param object[] $reflections
      * @return object[]
@@ -48,11 +62,13 @@ final class TransformerCollector
         }
 
         // @todo: sort here!, before ElementSorter
-        //if ($sort) { FIXME
+        $matchingTransformer = $this->detectMatchingTransformer($reflections);
+
+        if ($matchingTransformer instanceof SortableTransformerInterface) {
             uasort($elements, function ($firstElement, $secondElement) {
                return strcmp($firstElement->getName(), $secondElement->getName());
             });
-        //}
+        }
 
         return $elements;
     }
--- a/packages/Reflection/src/TransformerCollector.php
+++ b/packages/Reflection/src/TransformerCollector.php
@@ -32,6 +32,9 @@ final class TransformerCollector
         $this->transformers[] = $transformer;
     }
 
+    /**
+     * @param object[] $reflections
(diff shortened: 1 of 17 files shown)


packages/ThemeDefault/src/function.latte
```diff
@@ -49,7 +49,7 @@
                 </td>
                 <td class="value"><code>{block|strip}
                     <span class="property-name">{if $parameter->isPassedByReference()}&amp; {/if}${$parameter->getName()}</span>
-                    {if $parameter->isDefaultValueAvailable()} = {$parameter->getDefaultValueDefinition()|highlightValue:$function|noescape}{elseif $parameter->isVariadic()},…{/if}
+                    {if $parameter->isDefaultValueAvailable()} = {$parameter->getDefaultValueDefinition()|phpHighlight:$function|noescape}{elseif $parameter->isVariadic()},…{/if}
```
TomasVotruba: Did you find any other filter name miss-matches? If so, you could fix them all.

packages/ThemeDefault/src/function.latte
```diff
@@ -36,7 +36,7 @@
 
     {if count($function->getParameters())}
         <table class="summary table table-responsive table-bordered table-striped" id="parameters">
-            <tr><th>Parameters Summary</th></tr>
+            <tr><th colspan="3">Parameters Summary</th></tr>
```
TomasVotruba: :+1:

packages/Reflection/src/TransformerCollector.php
```diff
@@ -48,9 +48,11 @@ public function transformGroup(array $reflections): array
...
 
         // @todo: sort here!, before ElementSorter
-        uasort($elements, function ($firstElement, $secondElement) {
-           return strcmp($firstElement->getName(), $secondElement->getName());
-        });
+        if ($sort) {
+            uasort($elements, function ($firstElement, $secondElement) {
+               return strcmp($firstElement->getName(), $secondElement->getName());
+            });
+        }
```
TomasVotruba: I wonder what cases are useful to sort.
Even methods and constants should be in 1:1 to original, so they're easier to find. 
IMO, sorting is matter of coding standard, not Apigen.

Classes, traits and interfaces make sense to me. What do you think?

+ Is this the best place to sort such elements?

TomasVotruba:
That was not intention, you're right.

packages/ThemeDefault/src/function.latte
vlastavesely: :( it is also in `interface.latte`. Is there any way to edit particular commit from github directly?

packages/Reflection/src/TransformerCollector.php
vlastavesely: Hm, probably yes, the methods and constants should not be sorted too. Do you think it could be removed totally?

I do not think it belongs in here.

packages/ThemeDefault/src/function.latte
TomasVotruba: Great!
No need for squashing commits here, just add a new one. I squash it to 1 on merge anyway.

packages/Reflection/src/TransformerCollector.php
TomasVotruba: > Hm, probably yes, the methods and constants should not be sorted too. Do you think it could be removed totally?

Yes, I agree we should remove it. It should be placed somewhere else.

Do you have idea where to put it? 
I think that only elements that require sorting are:

- classes
- traits
- interfaces
- exceptions
- all those in namespaces

In particular Generators and `$all<>`.

vlastavesely: > In particular Generators and `$all<>` properties.

You mean this code in `ElementsTemplateVariablesEventSubscriber`?
```php
'allClasses' => $this->reflectionStorage->getClassReflections()
```
This is not modified by the `TransformerCollector`, it is a `ReflectionStorage` thing.

The sorting is - according to me - needed/useful also here:

```
ClassReflection::getInterfaces()
ClassReflection::getOwnInterfaces()
ClassReflection::getTraits()

InterfaceReflection::getInterfaces()
ClassReflection::getOwnInterfaces()

Parser::transformBetterFunctionReflections()
Parser::transformBetterClassInterfaceAndTraitReflections()
```

What about to leave the sorting where it is but to invert ins behavior? By default disable it and only when second parameter is `TRUE`, allow it.

TomasVotruba: > The sorting is - according to me - needed/useful also here:

Thanks for the list. It's nice to have overview.

> What about to leave the sorting where it is but to invert ins behavior? By default disable it and only when second parameter is TRUE, allow it.

**I agree. I'd go for the place**, just ommit manuall parameter (human factor/responsibility). Someone might add `get<some>Interfaces()` method in the future and could skip the parameter.

I'd check for the type (`in_array($currentType, $sortableTypes)` in the `transformGroup()` method. But that would break encapsulation, since `TransformerCollector` should have no knowledge about specific `Transformer`s. 

Mmm...

### The best I can think of would be

1. introduce `SortableTransformerInterface` in [`Contract/Transformer`](https://github.com/ApiGen/ApiGen/tree/d302650ae99fc796e223831131afce231def825f/packages/Reflection/src/Contract/Transformer) directory.

2. Detect matching tranformer - [this can be moved to standalone method](https://github.com/ApiGen/ApiGen/blob/d302650ae99fc796e223831131afce231def825f/packages/Reflection/src/TransformerCollector.php#L65-L67).

3. Sort elements if matching transformer implements this interface:

```php
$matchingTransformer = $this->detectMatchingTransformer($reflections);

if ($matchingTransformer instanceof SortableTransformerInterface) {
	uasort($elements, function ($firstElement, $secondElement) {
	   return strcmp($firstElement->getName(), $secondElement->getName());
	});	
}


return $elements;
```

4. Add `SortableTransformerReflection` to Transformers we need to sort:

```
- ClassReflectionTransformer
- InterfaceReflectionTransformer
- TraitReflectionTransformer
- FunctionReflectionTransformer
``` 


This way it's solid, having one place to extend with new transformer and hard to mess up (not impossible, but hard :smiley:  ).


What do you think?

vlastavesely:
I think it is worth a try. :laughing:

vlastavesely:
Would not it be better to simply ask whether class of `$reflections` can be sorted?
Just `if ($reflections[0] instanceof Something)` and allow the sorting for classes, interfaces etc...?

The checking of the matching transformer seems to me to be a little bit redundant :)

TomasVotruba:
That would be simpler at the moment, but break open-closed principle.

Collector should not be aware of specific items it contains.

I've found [very nice and short example](https://github.com/wataridori/solid-php-example/blob/b84657cb736f86dda1453061d15df01f260e5140/2-open-closed-principle.php#L24-L28), that explains this.

packages/Reflection/src/TransformerCollector.php
```diff
@@ -48,11 +62,13 @@ public function transformGroup(array $reflections): array
         }
 
         // @todo: sort here!, before ElementSorter
```
TomasVotruba: This note can be remove as well not.

TomasVotruba:
BEAUTIFUL! :100:

TomasVotruba:
Thank you for solid work :+1:  Merged.

PR #939, opened 2017-07-08
strike out deprecated classes, interfaces, etc. in the main summaries

vlastavesely:
In namespace summary it is already striked out, in the main summaries it was not.

deprecated


TomasVotruba:
Nice catch, thanks!

PR #943, opened 2017-07-08
:art: Minor improvement in 'finished' msg.

nelson6e65:
Use `info` context colour for `Finished!`, like you can see in https://travis-ci.org/ApiGen/ApiGen/builds/251577795#L727 log (in line 728).


TomasVotruba:
Why not :)

Issue #930, opened 2017-07-05
Add support for file source


POPSuL:
Now we can't provide file as source, and...

```
% php ./bin/apigen generate -vvv --destination=/var/www/api /var/www/core/src /var/www/core/system.php /var/www/core/exceptions.php

                                                                     
  [Roave\BetterReflection\SourceLocator\Exception\InvalidDirectory]  
  "/var/www/core/system.php" must be a directory, not a file         
                                                                     

Exception trace:
 () at /home/popsul/projects/ApiGen/vendor/roave/better-reflection/src/SourceLocator/Exception/InvalidDirectory.php:18
 Roave\BetterReflection\SourceLocator\Exception\InvalidDirectory::fromNonDirectory() at /home/popsul/projects/ApiGen/vendor/roave/better-reflection/src/SourceLocator/Type/DirectoriesSourceLocator.php:39
 Roave\BetterReflection\SourceLocator\Type\DirectoriesSourceLocator->Roave\BetterReflection\SourceLocator\Type\{closure}() at n/a:n/a
 array_map() at /home/popsul/projects/ApiGen/vendor/roave/better-reflection/src/SourceLocator/Type/DirectoriesSourceLocator.php:47
 Roave\BetterReflection\SourceLocator\Type\DirectoriesSourceLocator->__construct() at /home/popsul/projects/ApiGen/packages/Reflection/src/Parser/Parser.php:121
 ApiGen\Reflection\Parser\Parser->createDirectoriesSource() at /home/popsul/projects/ApiGen/packages/Reflection/src/Parser/Parser.php:52
 ApiGen\Reflection\Parser\Parser->parseDirectories() at /home/popsul/projects/ApiGen/src/Application/ApiGenApplication.php:55
 ApiGen\Application\ApiGenApplication->runCommand() at /home/popsul/projects/ApiGen/src/Console/Command/GenerateCommand.php:61
 ApiGen\Console\Command\GenerateCommand->execute() at /home/popsul/projects/ApiGen/vendor/symfony/console/Command/Command.php:264
 Symfony\Component\Console\Command\Command->run() at /home/popsul/projects/ApiGen/vendor/symfony/console/Application.php:869
 Symfony\Component\Console\Application->doRunCommand() at /home/popsul/projects/ApiGen/vendor/symfony/console/Application.php:223
 Symfony\Component\Console\Application->doRun() at /home/popsul/projects/ApiGen/vendor/symfony/console/Application.php:130
[...]

PR #944, opened 2017-07-11
added support for FileLocator

vlastavesely:
#930


vlastavesely:
:question: What was meant by
> update SourceOption

There is "Dirs or files documentation is generated for" in description, so it should be OK.

packages/Reflection/tests/Parser/FileLocatorTest.php
```diff
@@ -0,0 +1,34 @@
...
+        $parser = $this->container->get(Parser::class);
+
+        $parser->parseFilesAndDirectories([
+            __DIR__ . '/Source/ClassWithParentFromAnotherSource.php',
+            __DIR__ . '/AnotherSource/ParentClassFromAnotherSource.php'
+        ]);
+
+        $this->assertTrue(class_exists(ClassWithParentFromAnotherSource::class));
+        $this->assertTrue(class_exists(ParentClassFromAnotherSource::class));
+    }
```
TomasVotruba: `class_exists()` probably doesn't test anything related to parser.

Here would be better use `ReflectionStorage` services and tests `getClasses()` that contains these classes.

```php
$classReflections = $reflectionStoraget->getClassReflections();
$this->assertArrayHasKey(ClassWithParentFromAnotherSource:class, $classReflections);
```

packages/Reflection/tests/Parser/FileLocatorTest.php
```diff
@@ -0,0 +1,34 @@
...
+
+    public function testMultiple(): void
+    {
+        $parser = $this->container->get(Parser::class);
+
+        $parser->parseFilesAndDirectories([
+            __DIR__ . '/Source/ClassWithParentFromAnotherSource.php',
+            __DIR__ . '/AnotherSource/ParentClassFromAnotherSource.php'
+        ]);
```
TomasVotruba: Here would make sense to use 1 test method - single class and single directory. That way it will be covered all.

packages/Reflection/src/Parser/Parser.php
```diff
@@ -133,4 +165,17 @@ private function createDirectoriesSource(array $directories): SourceLocator
...
+     */
+    private function createFilesSource(array $files): SourceLocator
+    {
+        $locators = [];
+        foreach ($files as $file) {
+            $locators[] = new SingleFileSourceLocator($file);
+        }
+
+        return new AggregateSourceLocator($locators);
+    }
```
TomasVotruba: Beautiful work :clap:

packages/Reflection/src/Parser/Parser.php
```diff
@@ -46,9 +47,27 @@ public function __construct(
     }
 
     /**
+     * @param string[] $sources
+     */
+    public function parseFilesAndDirectories(array $sources): void
+    {
+        $files = $directories = [];
```
TomasVotruba: Could you split this into 2 lines please? For better readability and current coding standard convention.

TomasVotruba:
> There is "Dirs or files documentation is generated for" in description, so it should be OK.

Oh, that was probably future-proof :) allright then.

packages/Reflection/tests/Parser/FileLocatorTest.php
vlastavesely: That is not only in this test :(

TomasVotruba: I see.

There is no need for `Parser/FileLocatorTest.php`. It's too small and tests one publich method of `ParserTest`.

`ParserTest` should cover all its options.

vlastavesely: For example:

https://github.com/ApiGen/ApiGen/blob/master/packages/Reflection/tests/Parser/ExtendingClassTest.php
https://github.com/ApiGen/ApiGen/blob/master/packages/Reflection/tests/Parser/UseDifferentVendorTest.php

TomasVotruba: In `UseDifferentVendor` it might make sense. I didn't have the use case on my own, so I don't know what exactly was the issue and how to confirm it.

In `ExtendingClassTest` it could make sense to test `ReflectionStorage` for presence of parent class.

vlastavesely: OK.
Is `ParserTest` somewhere or do you mean to create new? I do not see any.

TomasVotruba: I thought there already is. 

> ...do you mean to create new?

Yes please. You can rename this test: `FileLocatorTest` to `ParserTest`.

TomasVotruba:
Your work has been merged, good job

PR #946, opened 2017-07-12
added highlighting of method's param default value

vlastavesely:
screenshot-3

:question:  any reason why default color of the `code` is reddish in color? It is unintended leak from the Bootstrap?

### DIFF
--- a/packages/ThemeDefault/src/partial/method.latte
+++ b/packages/ThemeDefault/src/partial/method.latte
@@ -31,7 +31,7 @@
                             {/if}
                             <span class="property-name">{if $parameter->isPassedByReference()}&amp; {/if}${$parameter->getName()}</span>
 
-                        {if $parameter->getDefaultValueDefinition()} = {$parameter->getDefaultValueDefinition()|phpHighlight|noescape}{elseif $parameter->isVariadic()},&hellip;{/if}</span>{sep}, {/sep}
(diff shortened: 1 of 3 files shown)


TomasVotruba:
Thanks for the PR! I'm curious. How did it look before? :)

> :question: any reason why default color of the code is reddish in color? It is unintended leak from the Bootstrap?

Probably. Which one is "normal"?

Also, could you keep "…" or did that cause any troubles? It's more readable and nowadays UTF-8 is supported by default.

vlastavesely:
> How did it look before? :)

All was reddish without bold. Do you need a screenshot? :camera_flash:  

>  Which one is "normal"?

Normal would be black and the reddish color could be used only when special class is added to `code` - for example for the `public`, `static` keywords etc. But not in the list of parameters - there it looks quite strange. Do you agree?

vlastavesely:
old

It looked like this - this is a screenshot from older PR.

TomasVotruba:
Yea, it makes sense. I agree! 🌳 

Thanks for screen, I understand now.

vlastavesely:
screenshot-4

Looks much better now! But I am not sure with the class name - there can be even another things than keywords. Any tip for better name?

TomasVotruba:
I'm fine with "keyword", makes sense

vlastavesely:
> I'm fine with "keyword", makes sense

:+1:

PR #950, opened 2017-07-13
Fixed link generation for type hints.

vlastavesely:
I just have started to work on the function return types I promised. But I have found another issue that is connected to it.

There was a problem that functions `getTypeHintClassOrInterfaceReflection()` relied on `class_exists()`. But it could happen that the class is not autoloaded before BetterReflection tries to locate it. And that is problem: for classes that were not located yet, it is going to fail. And for internal PHP classes it is going to work and generate a link to page that will not be generated.

I think that it would be better to check existence of the class it in the `ReflectionStorage` directly. I do not think that is is responsibility of the `ClassReflection` to know about other classes, it should be a matter of `ReflectionStorage`. And also there should be generated only links leading to elements really present in the ReflectionStorage (otherwise it could generate dead links to missing pages) so it is another reason to change it in a way I propose.

Additionally, there was bug with generating links - it printed just the path to the file but not generate a real link.

Originally, it looked like this:
bad

And after the fixes:
good

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -63,6 +63,7 @@
             "packages/Reflection/tests/Reflection/Function_/FunctionParameterReflection/Source/functionWithConstant.php",
             "packages/Reflection/tests/Reflection/Function_/FunctionParameterReflection/Source/functionWithVariadic.php",
             "packages/Reflection/tests/Reflection/Function_/FunctionParameterReflection/Source/functionWithClass.php",
+            "packages/StringRouting/tests/Latte/Filter/Source/TestClass.php",
             "packages/Annotation/tests/AnnotationSubscriber/SeeAnnotationSubscriberSource/SomeClassWithSeeAnnotations.php"
         ]
     },
--- a/composer.json
+++ b/composer.json
@@ -63,7 +63,6 @@
             "packages/Reflection/tests/Reflection/Function_/FunctionParameterReflection/Source/functionWithConstant.php",
             "packages/Reflection/tests/Reflection/Function_/FunctionParameterReflection/Source/functionWithVariadic.php",
             "packages/Reflection/tests/Reflection/Function_/FunctionParameterReflection/Source/functionWithClass.php",
-            "packages/StringRouting/tests/Latte/Filter/Source/TestClass.php",
             "packages/Annotation/tests/AnnotationSubscriber/SeeAnnotationSubscriberSource/SomeClassWithSeeAnnotations.php"
         ]
     },
--- a/packages/StringRouting/src/Latte/Filter/StringRoutingFiltersProvider.php
+++ b/packages/StringRouting/src/Latte/Filter/StringRoutingFiltersProvider.php
@@ -58,7 +58,7 @@ final class StringRoutingFiltersProvider implements FilterProviderInterface
                 $reflection = $this->reflectionStorage->getClassOrInterface($className);
                 if ($reflection) {
                     $link = Html::el('a');
-                    $link->href = $this->router->buildRoute(ReflectionRoute::NAME, $reflection);
+                    $link->setAttribute('href', $this->router->buildRoute(ReflectionRoute::NAME, $reflection));
                     $link->setText($className);
                     return $link;
                 } else {
--- a/packages/StringRouting/src/Latte/Filter/StringRoutingFiltersProvider.php
+++ b/packages/StringRouting/src/Latte/Filter/StringRoutingFiltersProvider.php
@@ -4,11 +4,13 @@ namespace ApiGen\StringRouting\Latte\Filter;
 
 use ApiGen\Contract\Templating\FilterProviderInterface;
 use ApiGen\Reflection\Contract\Reflection\AbstractReflectionInterface;
+use ApiGen\Reflection\ReflectionStorage;
 use ApiGen\StringRouting\Route\NamespaceRoute;
 use ApiGen\StringRouting\Route\ReflectionRoute;
 use ApiGen\StringRouting\Route\SourceCodeRoute;
 use ApiGen\StringRouting\StringRouter;
 use Nette\InvalidArgumentException;
+use Nette\Utils\Html;
 
 final class StringRoutingFiltersProvider implements FilterProviderInterface
 {
@@ -17,9 +19,15 @@ final class StringRoutingFiltersProvider implements FilterProviderInterface
      */
     private $router;
 
-    public function __construct(StringRouter $router)
+    /**
+     * @var ReflectionStorage
+     */
+    private $reflectionStorage;
+
+    public function __construct(StringRouter $router, ReflectionStorage $reflectionStorage)
     {
         $this->router = $router;
+        $this->reflectionStorage = $reflectionStorage;
     }
 
     /**
@@ -43,6 +51,19 @@ final class StringRoutingFiltersProvider implements FilterProviderInterface
             'linkSource' => function ($reflection): string {
                 $this->ensureFilterArgumentsIsReflection($reflection, 'linkSource');
                 return $this->router->buildRoute(SourceCodeRoute::NAME, $reflection);
+            },
+
+            // use in .latte: {$className|buildLinkIfReflectionFound}
+            'buildLinkIfReflectionFound' => function (string $className) {
+                $reflection = $this->reflectionStorage->getClassOrInterface($className);
+                if ($reflection) {
+                    $link = Html::el('a');
+                    $link->href = $this->router->buildRoute(ReflectionRoute::NAME, $reflection);
+                    $link->setText($className);
+                    return $link;
+                } else {
+                    return $className;
+                }
             }
         ];
     }
--- a/packages/StringRouting/tests/Latte/Filter/Source/TestClass.php
+++ b/packages/StringRouting/tests/Latte/Filter/Source/TestClass.php
@@ -1,0 +1,7 @@
+<?php declare(strict_types=1);
+
+namespace TestNamespace;
+
+class TestClass
+{
+}
--- a/packages/StringRouting/tests/Latte/Filter/Source/TestClass.php
+++ b/packages/StringRouting/tests/Latte/Filter/Source/TestClass.php
@@ -1,6 +1,6 @@
 <?php declare(strict_types=1);
 
-namespace TestNamespace;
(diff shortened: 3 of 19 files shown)


vlastavesely:
I forgot fix tests :)

vlastavesely:
OK, tomorrow, I will split it a little bit. :bomb: 

You are right with the `ReflectionStorage` and templates. Do you propose some kinda wrapper over the `RS`? Do you have any meaningful name for it?

TomasVotruba:
I'd prefer having that in reflections, so they have all the api they might need. Instead of splitting it to 2 places.

vlastavesely:
One issue is that I think the checking that type hint is really being a class or not is not a responsibility of the reflection itself but of some container holding all the reflections.

And another one is that it might generate a link to an interface (or a class) that is not in the `ReflectionStorage` but actually exists in PHP. `Countable` or `Iterator` interfaces are quite good example. They exists according to `class_exists()` but they are not generated for `ReflectionStorage` knows nothing about them. And that is the problem. No matter whether they should be generated or not - if the `ReflectionStorage` does not know about them, there should not be any link generated. Don't you think?

TomasVotruba:
> No matter whether they should be generated or not - if the ReflectionStorage does not know about them, there should not be any link generated.

I agree with this. The name itself, e.g. 'Countable' should be returned then.

vlastavesely:
So how to achieve this? Inject the `ReflectionStorage` inside all reflections being generated inside the ApiGen? External translator of the class name to class reflection (using `ReflectionStorage`) seems better to me :bulb:

TomasVotruba:
That's seem a bit overkill. I think there is simpler way to do this.

Where exactly is entry point to this error? Could you share a link to line in app?

I think of simple change in filter, in pseudo code like:

```php
if (class_exists($class) && $this->isInternalClass($class)) {
    return $class;
}
```

vlastavesely:
Then you have to make a list of internal classes. And again, I do not think it is a good idea - for multiple versions of PHP with different internal classes. That is gonna fail for someone for sure.

Link to line in application? I just use a few of test classes but I have everything on another :computer:

vlastavesely:
OK, tomorrow, I will send you the code I use.

That reflection allows you to check whether it is user-defined is amazing, I will use it on another problem I found.

But, let me to think again :bulb:

Still, it can happen that `class_exists()` returns `TRUE` for some class that is autoloaded but actually not generated. For example, for some dependency that ApiGen shares with the code that the API is being generated for. Let us say that we are speaking about `Nette\Utils\FileSystem`. In is not an internal class but it exists since it has been autoloaded by ApiGen's composer autoloader. Then, your pseudocode generates link. But at the time, there are not generated reflections for dependencies. If I did not miss something, it is going to generate a dead link.

Am I right :question:

TomasVotruba:
You are right. For such case I think it should be generated. I'd expect as same behavior as from IDE. When parent class is in vendor, I should be able to click through.

To make this work, all method have to be called to instantiate those dependencies to `ReflectionStorage`.  They have to be loaded *before* generate section, not during it. 

This should be handled by [`ReflectionWarmupper`](https://github.com/ApiGen/ApiGen/blob/master/packages/Element/src/Cache/ReflectionWarmUpper.php). This method `getTypeHintClassOrInterfaceReflection()`  should be definitelly added there.

vlastavesely:
@TomasVotruba, could you test something? I just made a test whether BetterReflection does autoload a class or not. Install BetterReflection via compose, copy this files into the same directory and run `index.php`.

## index.php
```php
<?php

use Roave\BetterReflection\Reflector\ClassReflector;
use Roave\BetterReflection\SourceLocator\Type\SingleFileSourceLocator;

include __DIR__ . '/vendor/autoload.php';

$reflector = new ClassReflector(new SingleFileSourceLocator(__DIR__ . '/SomeClass.php'));
$classInfo = $reflector->reflect('SomeNamespace\SomeClass');
 $properties = $classInfo->getProperties();

// Just a test whether it really reflects the class
print $classInfo->getName() . ' has been reflected by BR' . "\n";
print '  and has ' . count($properties) . ' properties (' . implode(',', array_keys($properties)) . ')' . "\n";

// Manual loading
//include 'SomeClass.php';

if (class_exists('SomeNamespace\SomeClass')) {
	print "Yes, class has been autoloaded.\n";
} else {
	print "\e[31;1mHouston, we have a problem. Class has not been autoloaded.\e[0m\n";
}
```

## SomeClass.php
```php
<?php

namespace SomeNamespace;

class SomeClass
{
	public $prop = 11;
	private $priv = 1;
}
```

For me, class is not autoloaded.

vlastavesely:
Also, it explains the problem in #941

TomasVotruba:
Thanks for copy-paste setup, I've just tried it:

image


I think that is feature of BetterReflection: https://github.com/Roave/BetterReflection#why-is-it-better

> You can reflect on classes that are not already loaded, without loading them

That class should be loaded by composer, because some other parts like parent class you references, depend on classes that are not directly parsed by BetterReflection.

I think there should be exception "class is not loaded, you should register it in composer".

vlastavesely:
When you run `apigen/bin/apigen generate ../../path/to/source ../path/to/another/source.php`, composer CANNOT know about the classes. How could it?

It can know all classes only when you require `apigen/apigen` as dependency of a project and then run `vendor/bin/apigen generate app`.

TomasVotruba:
I think with `ComposerLocator` you've added. It would fail otherwise.

vlastavesely:
The `ComposerLocator` loads only files in .`./vendor` directory (relatively to source dir). But if you add for example `SingleFileLocator`, the class will not be autoloaded. :)

TomasVotruba:
I see. Some cases don't support this, but `vendor/autoload.php` is still preffered direction.

What do you suggest here?

vlastavesely:
Hmm, the only place where you can find an information about all the classes actually located, is the `ReflectionStorage`. I agree with you that passing whole `ReflectionStorage` is a violation of the SOLID so I think there should be passed a wrapper over RS containing just `getClassOrInterface()`, `getClass()` and `getInterface()` methods.

Or maybe, if `Latte` filters can have dependencies, introduce a filter that could take class name and return plain string when class does not exist in the `ReflectionStorage` or instance of `Nette\Utils\Html` representing a ready link when it does exist in the RS. In this case, there would not be need to pass any dependency into templates directly.

TomasVotruba:
> Or maybe, if Latte filters can have dependencies, introduce a filter that could take class name and return plain string when class does not exist in the ReflectionStorage or instance of Nette\Utils\Html representing a ready link when it does exist in the RS. In this case, there would not be need to pass any dependency into templates directly.

I'm :+1: on this. That was my suggestion here: https://github.com/ApiGen/ApiGen/pull/950#issuecomment-315166244, but I didn't typed it exactly right.

It would be in `StringRoutingFiltersProvider` here: https://github.com/ApiGen/ApiGen/blob/master/packages/StringRouting/src/Latte/Filter/StringRoutingFiltersProvider.php#L37

But that is already in link in template... 

```html
<a href="{$reflection|linkReflection}">{$reflection->getName()}</a>
```

**So new filter have to be added to `StringRoutingFiltersProvider` that would work as a wrapper:**

```html
{$element|buildLinkIfReflectionFound}
```

What do you think?

I would not use it instead of all `{$reflection|linkReflection}` cases, but only where sense.

What are exact places in [templates](https://github.com/ApiGen/ApiGen/blob/master/packages/ThemeDefault/src/), where is this needed?

vlastavesely:
> I would not use it instead of all {$reflection|linkReflection} cases, but only where sense.

I agree.

> What are exact places in templates, where is this needed?

It is needed in links on parameter typehints, property types and (in the future) on method return type hints.

TomasVotruba:
Ok. Do you have all you need to know?

vlastavesely:
I think so :+1:

vlastavesely:
As we agreed, I made the changes, still without tests. If you like it, I will fix it.

Looks still the same as on the first version:
s

TomasVotruba:
It looks very ellegant. I admire a code where a feature is added by deleting more code than added :bowing_man: 

All it needs is test for this filter - both cases - and I think it's ready for merge:
https://github.com/ApiGen/ApiGen/pull/950/files#diff-48b6678dccc5c6b98b4d81359c79c3caR57

packages/StringRouting/src/Latte/Filter/StringRoutingFiltersProvider.php
```diff
@@ -58,7 +58,7 @@ public function getFilters(): array
                 $reflection = $this->reflectionStorage->getClassOrInterface($className);
                 if ($reflection) {
                     $link = Html::el('a');
-                    $link->href = $this->router->buildRoute(ReflectionRoute::NAME, $reflection);
+                    $link->setAttribute('href', $this->router->buildRoute(ReflectionRoute::NAME, $reflection));
```
TomasVotruba: Nice!

composer.json
```diff
@@ -63,6 +63,7 @@
             "packages/Reflection/tests/Reflection/Function_/FunctionParameterReflection/Source/functionWithConstant.php",
             "packages/Reflection/tests/Reflection/Function_/FunctionParameterReflection/Source/functionWithVariadic.php",
             "packages/Reflection/tests/Reflection/Function_/FunctionParameterReflection/Source/functionWithClass.php",
+            "packages/StringRouting/tests/Latte/Filter/Source/TestClass.php",
```
TomasVotruba: This class should be autoloaded by PSR-4 for StringRouting, if possible. So it makes it easier to refactor and to find a class. It also prevent having class duplicities in various places.

Here should be only functions and classes that cannot be found in another way, e.g. class without namespace.

packages/StringRouting/tests/Latte/Filter/Source/TestClass.php
```diff
@@ -0,0 +1,7 @@
+<?php declare(strict_types=1);
+
+namespace TestNamespace;
+
+class TestClass
```
TomasVotruba: Could you please complete this class to respect PSR-4? See explanation above.

Thank you

TomasVotruba:
Just one namespace change and it's ready to merge :+1:

TomasVotruba:
Merged!

Thank you for your cooperation, I'm very happy with this solution.

PR #954, opened 2017-07-17
deprecate methods of deprecated interface

vlastavesely:
Done in the same way as in class or trait.


TomasVotruba:
:+1: Nice catch

PR #955, opened 2017-07-18
RelativePathResolver: fix for functions loaded from file source

vlastavesely:
When second parameter of `substr` is bigger than actual size of the string, it returns `FALSE` and it fails on type hint restriction where just a string is expected.


TomasVotruba:
Thank you

PR #956, opened 2017-07-18
Reflection: allowed nullable filenames

vlastavesely:
Output of `BetterReflection*::getFileName()` is nullable for internal PHP classes do not have source. We do not support generating API for PHP's internal classes but we need to. After it, current version would fail on type hint check.

### DIFF
--- a/packages/Reflection/src/Contract/Reflection/FileNameAwareReflectionInterface.php
+++ b/packages/Reflection/src/Contract/Reflection/FileNameAwareReflectionInterface.php
@@ -1,0 +1,8 @@
+<?php declare(strict_types=1);
+
+namespace ApiGen\Reflection\Contract\Reflection;
+
+interface FileNameAwareReflectionInterface
+{
+    public function getFileName(): ?string;
+}
--- a/packages/Reflection/src/Contract/Reflection/FileNameAwareReflectionInterface.php
+++ b/packages/Reflection/src/Contract/Reflection/FileNameAwareReflectionInterface.php
@@ -4,5 +4,8 @@ namespace ApiGen\Reflection\Contract\Reflection;
 
 interface FileNameAwareReflectionInterface
 {
+    /**
+     * @return string|null  returns null for PHP internal classes
+     */
     public function getFileName(): ?string;
 }
--- a/src/Console/Progress/StepCounter.php
+++ b/src/Console/Progress/StepCounter.php
@@ -4,7 +4,7 @@ namespace ApiGen\Console\Progress;
 
 use ApiGen\Element\Namespace_\ParentEmptyNamespacesResolver;
 use ApiGen\Element\ReflectionCollector\NamespaceReflectionCollector;
-use ApiGen\Reflection\Contract\Reflection\AbstractReflectionInterface;
+use ApiGen\Reflection\Contract\Reflection\FileNameAwareReflectionInterface;
 use ApiGen\Reflection\ReflectionStorage;
 
 final class StepCounter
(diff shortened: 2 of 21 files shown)


src/Console/Progress/StepCounter.php
```diff
@@ -52,11 +52,38 @@ public function getStepCount(): int
...
+            }
+        }
+
+        foreach ($this->reflectionStorage->getTraitReflections() as $traitReflection) {
+            if ($traitReflection->getFileName()) {
+                $count++;
+            }
+        }
+
+        foreach ($this->reflectionStorage->getFunctionReflections() as $functionReflection) {
```
TomasVotruba: I'd add helper method here, that would count it internally:

```php
$this->getCountForGeneratedReflections($this->reflectionStorage->getFunctionReflections());
```

vlastavesely: Good point :+1:

TomasVotruba:
Thanks for the update!


I wonder, where do you need internal classes apart their class name?

They were a huge code overhead in previous version, but not really useful and rather broken. I would not like to get to similar state, where they have own reflection api (InternalClassReflection, InternalClassMethodReflection etc.), generators and so on.

src/Console/Progress/StepCounter.php
```diff
@@ -50,42 +51,30 @@ public function getStepCount(): int
...
-        foreach ($this->reflectionStorage->getTraitReflections() as $traitReflection) {
-            if ($traitReflection->getFileName()) {
-                $count++;
-            }
-        }
-
-        foreach ($this->reflectionStorage->getFunctionReflections() as $functionReflection) {
-            if ($functionReflection->getFileName()) {
+        foreach ($reflections as $reflection) {
+            if ($reflection->getFileName()) {
```
TomasVotruba: This method probably doesn't exist on `AbstractReflectionInterface`: https://github.com/ApiGen/ApiGen/blob/22562701ce8b9174fcff2574cbb79752f20156cb/packages/Reflection/src/Contract/Reflection/AbstractReflectionInterface.php#L7

Adding this method might fail for partial reflections like `ClassMethodReflection`, because they don't really need it.

Maybe we could add `FileNameAwareInterface`?

src/Console/Progress/StepCounter.php
```diff
@@ -50,42 +51,30 @@ public function getStepCount(): int
...
                 $count++;
             }
         }
 
         return $count;
     }
 
+    private function getSourceCodeStepCount(): int
```
TomasVotruba: Private methods should be ordered by call time (earlier called should be first).
This method should be above `getSourceCodeCountForReflections()`, took me while to read they depend on each other :

vlastavesely:
>  I would not like to get to similar state, where they have own reflection api (InternalClassReflection, InternalClassMethodReflection etc.), generators and so on.

Not even me. If it would be as hard, I would to skip it also. But I tried to add all parent classes and all interfaces implemented into `ReflectionStorage` and generate the API. For internal classes, nothing was broken (on the ones I tested) except the links to source codes. That should not be problem to fix.

Anyway, for `BetterReflection's`functions can return `NULL` we should handle the possibility of it properly.

vlastavesely:
By the way, good idea with the `FileNameAwareReflectionInterface` 👍

packages/Reflection/src/Contract/Reflection/FileNameAwareReflectionInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php declare(strict_types=1);
+
+namespace ApiGen\Reflection\Contract\Reflection;
+
+interface FileNameAwareReflectionInterface
+{
+    public function getFileName(): ?string;
```
TomasVotruba: I'd add a note here, that it can be null for PHP internal classes. A message to next generations :)

TomasVotruba:
Ok, your words make sense to me. Thanks for the work :+

Is it ready to merge now?

vlastavesely:
> Is it ready to merge now?

I think so.

TomasVotruba:
Done :clapper: 

Thanks a lot!

PR #957, opened 2017-07-18
Parser: fixed file source loader

vlastavesely:
When loading only from file (without directories), autoloaded classes or PHP's internal interfaces were ignored.

Without the fix, test has been throwing this:

```
Roave\BetterReflection\Reflector\Exception\IdentifierNotFound:
Roave\BetterReflection\Reflection\ReflectionClass "Countable" could not be found in the located source
```


TomasVotruba:
Nice :+1:

PR #958, opened 2017-07-22
added support for interface constants [WIP]

vlastavesely:
**Waits on PR <s>I am going to make into `BetterReflection`</s> Roave/BetterReflection#292.**

Also, we should get rid of
```
{if $constant instanceof \ApiGen\Reflection\Contract\Reflection\Partial\AnnotationsInterface}{$constant|description}{/if}
 ```
in templates. It is not the first case (it can be found in `function.latte`) so I propose to merge it as it is and then we should find another solution - maybe introduce some `DescriptionAwareInterface` for the latte filter....?

### DIFF
--- a/packages/Reflection/src/Reflection/Class_/ClassConstantReflection.php
+++ b/packages/Reflection/src/Reflection/Class_/ClassConstantReflection.php
@@ -84,6 +84,11 @@ final class ClassConstantReflection implements ClassConstantReflectionInterface,
         return $this->constant->getValue();
     }
 
+    public function getValueDefinition(): string
+    {
+        return 'TODO'; // $this->constant->getValueAsString(); FIXME
+    }
+
     public function isDeprecated(): bool
     {
         if ($this->getDeclaringClass()->isDeprecated()) {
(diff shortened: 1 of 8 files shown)


TomasVotruba:
This can be merged, right?

vlastavesely:
Hm, it could be. Only hadling of default value rendering mudr be fixed after my other pull is revolved.

TomasVotruba:
Ok, shall I merge them both and then you'll prepare fix PR?

vlastavesely:
Yes, I think it could be merged. It will only output `'TODO'` strings as default values of constants (this PR) and string names of constants as default values in parameters (#959) so output will be interpreted as string: `method($param = 'self::CONST')`.

On fix I am already working but it is delaying. I am waiting for my another PR to be resolved in BR and also, I have to solve something in my work so there was not enough time.

TomasVotruba:
Thanks for description of current state, it makes it easy to orientate in these changes for me.

Alright, let's merge it now then, so you have clean plate to work with.

PR #959, opened 2017-07-22
added ValueDumper

vlastavesely:
Getting rid of BetterReflection's functions `getDefaultValueAsString()`. Their future is quite uncertain Roave/BetterReflection#292

### DIFF
--- a/packages/ThemeDefault/src/function.latte
+++ b/packages/ThemeDefault/src/function.latte
@@ -45,7 +45,7 @@
                 </td>
                 <td class="value"><code>{block|strip}
                     <span class="property-name">{if $parameter->isPassedByReference()}&amp; {/if}${$parameter->getName()}</span>
-                    {if $parameter->isDefaultValueAvailable()} = {$parameter->getDefaultValueDefinition()|phpHighlight:$function|noescape}{elseif $parameter->isVariadic()},…{/if}
+                    {if $parameter->isDefaultValueAvailable()} = {$parameter->getDefaultValue()|dumpDefaultValue|phpHighlight:$function|noescape}{elseif $parameter->isVariadic()},…{/if}
                 {/block}</code></td>
                 <td>{if $parameter instanceof \ApiGen\Reflection\Contract\Reflection\Partial\AnnotationsInterface}{$parameter|description}{/if}</td>
             </tr>
--- a/packages/ThemeDefault/src/function.latte
+++ b/packages/ThemeDefault/src/function.latte
@@ -45,7 +45,7 @@
                 </td>
                 <td class="value"><code>{block|strip}
                     <span class="property-name">{if $parameter->isPassedByReference()}&amp; {/if}${$parameter->getName()}</span>
-                    {if $parameter->isDefaultValueAvailable()} = {$parameter->getDefaultValue()|dumpDefaultValue|phpHighlight:$function|noescape}{elseif $parameter->isVariadic()},…{/if}
+                    {if $parameter->isDefaultValueAvailable()} = {$parameter->getDefaultValue()|dumpDefaultValue|phpHighlight|noescape}{elseif $parameter->isVariadic()},…{/if}
                 {/block}</code></td>
                 <td>{if $parameter instanceof \ApiGen\Reflection\Contract\Reflection\Partial\AnnotationsInterface}{$parameter|description}{/if}</td>
             </tr>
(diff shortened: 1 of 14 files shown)


TomasVotruba:
Thanks for being up-to-date with current BetterReflection deprecations.

**Could you maybe move this to filter where this method is now used?**
 I'd like to avoid static where there is a another way to solve this.

vlastavesely:
What about this?

------
There is a known bug. Consider this:

```php

define('CONST', 'aa');

class SomeClass
{
	const CONST = 'bb';

	public function func($param = self::CONST);
}
```

Before this, ApiGen generated:

```
public function func($param = CONST);
```

and that is wrong for, in fact, there is default value 'aa' but in the code is 'bb'. After removing `getDefaultValueDefinition()` it returns correct default value but it would be better fo return something that would be rendered into `self::CONST`. Maybe instance of reflection of the constant? (Default value cannot be a class so we could use it to our advantage and make serializer to serialize reflections in way we want....)

Anyway, one bug or another, this could be solved separately.

TomasVotruba:
> Maybe instance of reflection of the constant?

This makes sense to me. I think previous approach was similar.

> Anyway, one bug or another, this could be solved separately.

:+1:

vlastavesely:
Added forgotten ValueDumper :) Now it should be done. Some tests (dumping constand as parameter's default value) had to be commented out for it cannot be fixed from here, it is another feature.

TomasVotruba:
Looks fine. Could add test for `DefaultValueDumper` please?

> Some tests (dumping constand as parameter's default value) had to be commented out for it cannot be fixed from here.

So you'll fix that in another PR? 
I'd merge them both when ready, so there is nothing extra broken in between them.

vlastavesely:
Test added.

```
So you'll fix that in another PR?
```

Yes, I will try. :) But for that, changes from this PR will be needed. How do we do that?

TomasVotruba:
> Yes, I will try. :) But for that, changes from this PR will be needed. How do we do that?

Great :) well, I usually create a new branch from this one and send PR from that branch to this one.
After accepting them just merge the never branch first.

packages/ThemeDefault/src/function.latte
```diff
@@ -45,7 +45,7 @@
                 </td>
                 <td class="value"><code>{block|strip}
                     <span class="property-name">{if $parameter->isPassedByReference()}&amp; {/if}${$parameter->getName()}</span>
-                    {if $parameter->isDefaultValueAvailable()} = {$parameter->getDefaultValueDefinition()|phpHighlight:$function|noescape}{elseif $parameter->isVariadic()},…{/if}
+                    {if $parameter->isDefaultValueAvailable()} = {$parameter->getDefaultValue()|dumpDefaultValue|phpHighlight:$function|noescape}{elseif $parameter->isVariadic()},…{/if}
                 {/block}</code></td>
```
TomasVotruba: I think `:$function` can be removed. Probably some legacy code.

vlastavesely:
> ...send PR from that branch to this one...

Like any other PR?

TomasVotruba:
@vlastavesely Yeah. When you create a PR, Github offers you to pick *target branch*. There you'll have `master` by default, but you!ll select this one instead.

vlastavesely:
Ugh, in BetterReflection, the changes got merged already. That means that with newly installed dependencies, people will not be able run ApiGen without this changes merged :(

TomasVotruba:
> Ugh, in BetterReflection, the changes got merged already. That means that with newly installed dependencies, people will not be able run ApiGen without this changes merged :(

We can avoid this by using specific commits in a `composer.json`.
Like this: https://github.com/Symplify/Symplify/blob/93615d7f1c6dad5d5e88a33b66a908b5fc0e9389/composer.json#L46

TomasVotruba:
@vlastavesely Vlasta, do you have some contact - website, email, FB or WhatsApp? I'd like to talk with you in some private (instant) chanel

vlastavesely:
@TomasVotruba Check info (at) tomasvotruba (dot) cz

vlastavesely:
@TomasVotruba I am getting some troubles with creating a new PR, I am not sure what settings of target branch should be used :)

For now, I have just one commit and it is still WIP. There will be need for some changes in BetterReflection, for now I am not sure what solution would be the best.

I also fixed another problem in BR - Roave/BetterReflection#296

TomasVotruba:
@vlastavesely Sure, no problem with branches.

I'll write you next week :)

TomasVotruba:
Thank you!

PR #960, opened 2017-07-23
typo

vlastavesely:
`constant` -> `betterConstantReflection`

1. Just "constant" is a little bit confusing in the code. I had to scroll to the top to see what it really contains.

2. In other places, the variable is named `better<Something>Reflection` so with this change it would be more consistent.


TomasVotruba:
Agreed!

Issue #942, opened 2017-07-08
Does not generate API for different vendor classes


vlastavesely:
In #916 we have fixed failing of the generator due to classes that are not autoloaded. After the #916, this is working well. But still, ApiGen does not generate API for the parent classes.

If I have class,
```php

class MyUser extends Nette\Security\User
{
}
```

API is going to be generated for the `MyUser` class. It also contains links to `Nette\Security\User`, but the files the links lead to are missing. The same problem is with missing interfaces.

**I have started to work on that** but it is quite complex task and it is going to take some time.

ITPrism:
Do you have any good news about this issue?

vlastavesely:
My work I have been working on, waits for sonething else. So, I could look at this issue tomorrow.

vlastavesely:
Closed by #971

PR #971, opened 2017-08-03
Resolving parent classes from other dirs/php internal classes [WIP]

vlastavesely:
Just a concept for a review. :construction: 

Tests are failing but it generates what I expected. :ok_hand: 

Related to #942


packages/Reflection/src/Parser/Parser.php
```diff
@@ -127,11 +127,44 @@ private function transformBetterFunctionReflections(FunctionReflector $functionR
      */
     private function transformBetterClassInterfaceAndTraitReflections(ClassReflector $classReflector): array
     {
-        $betterClassReflections = $classReflector->getAllClasses();
+        $betterClassReflections = $this->resolveParentClassesAndInterfaces($classReflector->getAllClasses());
```
TomasVotruba: I'd split this into 2 lines.

packages/Reflection/src/Parser/Parser.php
```diff
@@ -127,11 +127,44 @@ private function transformBetterFunctionReflections(FunctionReflector $functionR
...
 
         return $this->transformerCollector->transformGroup($betterClassReflections);
     }
 
+    /**
+     * @param ReflectionClass[]
+     */
+    private function resolveParentClassesAndInterfaces(array $betterClassReflections): array
```
TomasVotruba: Include "Traits" in the method name.

packages/Reflection/src/Parser/Parser.php
```diff
@@ -127,11 +127,44 @@ private function transformBetterFunctionReflections(FunctionReflector $functionR
...
+    {
+        $allReflections = [];
+
+        foreach ($betterClassReflections as $class) {
+            $allReflections[$class->getName()] = $class;
+
+            foreach ($class->getInterfaces() as $interface) {
```
TomasVotruba: I'd split these into 3 methods.

packages/Reflection/src/Parser/Parser.php
```diff
@@ -127,11 +127,44 @@ private function transformBetterFunctionReflections(FunctionReflector $functionR
...
 
+    /**
+     * @param ReflectionClass[]
+     */
+    private function resolveParentClassesAndInterfaces(array $betterClassReflections): array
+    {
+        $allReflections = [];
+
+        foreach ($betterClassReflections as $class) {
+            $allReflections[$class->getName()] = $class;
```
TomasVotruba: What about removing this line and using directorly `$betterClassReflections`? It includes already all found classes.

TomasVotruba:
Just few details, otherwise it looks good :+1: 

Could you add tests for final version that fail before this fix? So we know this fixes it. E.g. similar use case #942

packages/Reflection/src/Parser/Parser.php
vlastavesely: It includes but the items are identified by numeric index and not class name. So duplicities can occur.

TomasVotruba: I see. It makes sense then.

vlastavesely:
The `TraitReflectionTest` is failing for BR cannot find `ParentTrait` in `SimpleTrait`. No test does any assertion on it so it could be removed. But it would be better to find source of problems and fix it properly. 

@TomasVotruba could you take a look on it? You may see something I missed.

vlastavesely:
It is the Roave/BetterReflection#274 I think.

TomasVotruba:
@vlastavesely I recall this issue now. You can either remove it or send failing PR, that @asgrim mentions in [BetterReflection comment](https://github.com/Roave/BetterReflection/pull/274#issuecomment-317268219) into `phpDocumentor/TypeResolver` package.

vlastavesely:
I have a few things in process, so I should not start with another :) I commented it out so it can be easily added back.

TomasVotruba:
:+1: thank you

PR #975, opened 2017-08-09
Typehints refactoring [WIP]

vlastavesely:
Allow multiple typehints as array. At the time, it does not support indexed parameters, it requires a name:

```php
/**
 * @param int $foo
 */
function a($foo)
{
}
```

### DIFF
--- a/packages/ThemeDefault/src/partial/property.latte
+++ b/packages/ThemeDefault/src/partial/property.latte
@@ -3,9 +3,7 @@
         <code class="keyword">
             {if $property->isProtected()}protected{elseif $property->isPrivate()}private{else}public{/if} {if $property->isStatic()}static{/if}
 
-            {if count($property->getTypeHints())}
-                {foreach $property->getTypeHints() as $typehint}{$typehint|buildLinkIfReflectionFound}{sep}|{/sep}{/foreach}
-            {/if}
+            {foreach $property->getTypeHints() as $typehint}{$typehint|buildLinkIfReflectionFound}{sep}|{/sep}{/foreach}
         </code>
     </td>
 
--- a/packages/ThemeDefault/src/partial/property.latte
+++ b/packages/ThemeDefault/src/partial/property.latte
@@ -3,7 +3,9 @@
         <code class="keyword">
             {if $property->isProtected()}protected{elseif $property->isPrivate()}private{else}public{/if} {if $property->isStatic()}static{/if}
 
-            {$property->getTypeHint()|buildLinkIfReflectionFound}
+            {if count($property->getTypeHints())}
+                {foreach $property->getTypeHints() as $typehint}{$typehint|buildLinkIfReflectionFound}{sep}|{/sep}{/foreach}
+            {/if}
         </code>
     </td>
 
--- a/packages/Reflection/src/Contract/Reflection/AbstractParameterReflectionInterface.php
+++ b/packages/Reflection/src/Contract/Reflection/AbstractParameterReflectionInterface.php
@@ -4,7 +4,7 @@ namespace ApiGen\Reflection\Contract\Reflection;
 
 interface AbstractParameterReflectionInterface extends AbstractReflectionInterface
 {
-    public function getTypeHint(): string;
+    public function getTypeHints(): array;
 
     public function isDefaultValueAvailable(): bool;
 
--- a/packages/Reflection/src/Contract/Reflection/AbstractParameterReflectionInterface.php
+++ b/packages/Reflection/src/Contract/Reflection/AbstractParameterReflectionInterface.php
@@ -4,6 +4,9 @@ namespace ApiGen\Reflection\Contract\Reflection;
 
 interface AbstractParameterReflectionInterface extends AbstractReflectionInterface
 {
+    /**
+     * @return string[]
+     */
     public function getTypeHints(): array;
 
     public function isDefaultValueAvailable(): bool;
--- a/packages/Reflection/src/Contract/Reflection/Class_/ClassPropertyReflectionInterface.php
+++ b/packages/Reflection/src/Contract/Reflection/Class_/ClassPropertyReflectionInterface.php
@@ -16,7 +16,10 @@ interface ClassPropertyReflectionInterface extends AbstractClassElementInterface
      */
     public function getDefaultValue();
 
-    public function getTypeHint(): string;
+    /**
+     * @return string[]
+     */
+    public function getTypeHints(): array;
 
     public function getNamespaceName(): string;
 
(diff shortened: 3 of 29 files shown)


vlastavesely:
Proposed  Roave/BetterReflection#299

packages/ThemeDefault/src/partial/property.latte
```diff
@@ -3,7 +3,9 @@
         <code class="keyword">
             {if $property->isProtected()}protected{elseif $property->isPrivate()}private{else}public{/if} {if $property->isStatic()}static{/if}
 
-            {$property->getTypeHint()|buildLinkIfReflectionFound}
+            {if count($property->getTypeHints())}
+                {foreach $property->getTypeHints() as $typehint}{$typehint|buildLinkIfReflectionFound}{sep}|{/sep}{/foreach}
```
TomasVotruba: Since `getTypeHints()` returns always `array`, there is no need for condition `{if count($property->getTypeHints())}`.

If there is empty array (`array[]`), nothing will de shown. Same for `method.latte`

TomasVotruba:
Beautiful work :+1:

vlastavesely:
And I saw unnamed annotations more times than named ones.

I will check whether there is some easy way to go. If so, I will add support for it. Even my projects use unnamed `@param` annotations :)

TomasVotruba:
> Even my projects use unnamed @param annotations :)

I see, I just want to remind ApiGen is PHP 7.1 project that focuses on modern code bases. That's main advantage to other projects, that doesn't support PHP 7.1 or even PHP 7.0 in some cases.

What would be advantage of this?

```php
/** 
 * @param string
 * @param int
 * @param string[]
 * @param bool
 */
public function ($arg1, $arg2, $arg3, $arg4)
```

over this

```php
/** 
 * @param string[] $arg3
 */
public function (string $arg1, int $arg2, array $arg3, bool $arg4)
```

"Don't make programmer (and the code) think."


But if you still need it, feel free to implement it.

vlastavesely:
I decided to change doc comments in my projects to be `PSR-5` compliant. But I am afraid that missing support of indexed `@param` would be reported as issue soon...

vlastavesely:
@TomasVotruba what is wrong here: https://travis-ci.org/ApiGen/ApiGen/jobs/263060191 ?

I see just 152 lines of installation of certificates but nothing more.

TomasVotruba:
@vlastavesely There are some CS issues: https://travis-ci.org/ApiGen/ApiGen/jobs/263060191#L544

vlastavesely:
Strange, now I see it too. But the log is something different from what I saw....

TomasVotruba:
Yea, sometimes Travis stucks :) I usually give it a few minutes or restard the build.

vlastavesely:
A way opened here to simplify everything when `DocBlockFactory` creates type proper context.

Right now it breaks tests, for parameters type hints are rendered with full namespace. (See https://travis-ci.org/ApiGen/ApiGen/builds/263819712#L512) That is needed for generating links and so on.

But as link's anchor, the classes from the same namespace as the parent class is could be rendered without full namespace name. @TomasVotruba what do you think?

P.S. You yourself have already implemented indexed parameters, for the annotations are retrieved by param's position (https://github.com/ApiGen/ApiGen/blob/master/packages/Reflection/src/Reflection/AbstractParameterReflection.php#L107-L116). There was just problem with bad types resolving.

TomasVotruba:
I agree to keep the way it was before, just display element name without full namespace.

vlastavesely:
Only for classes of the same NS?

TomasVotruba:
I'd just used what was in the original code.  If there is fqn, use it, if short, use that.

vlastavesely:
> I'd just used what was in the original code. If there is fqn, use it, if short, use that.

I've been thinking about it and tested some scenarios. I am pretty much sure that the documentation and real code are two different things. Consider this:

```php

namespace TestNamespace;

use DifferentNamespace\DifferentClass;

class TestClass
{
	/**
	 * @param DifferentClass $class
	 */
	public function method($class)
	{
	}
}
```

For there are no `use` in documentation, using just `DifferentClass` like in the original code would be confusing. It would seem that it is `\TestNamespace\DifferentClass` while it is not. It can't be used what was in the original code.

I am wondering what solution would be the best - FQCN for another-namespace classes and CN for this-namespace classes? Or CN for all classes with tooltip with FQCN shown on mouse hover event, maybe with some icon indicating that the class belogngs into another NS?

-----

And another thing - I have replaced strings with `phpDocumetor`'s `Type` classes but not commited yet. It simplifies things even more but I am not sure about it. What do you think about using it instead of strings?

TomasVotruba:
> For there are no use in documentation, using just DifferentClass like in the original code would be confusing. It would seem that it is \TestNamespace\DifferentClass while it is not. It can't be used what was in the original code.

Oh, I see what you mean now. I got lost in the words. In anoher namespace, you could use FQCN then, to me it clear where the class is from.

> It simplifies things even more but I am not sure about it.

Let's give it a try then :wink:

TomasVotruba:
Nice work! The Type class approach looks very fancy :+1:

vlastavesely:
I am glad to read you like it. :heart:
Today, I have been on a trip so I did not work on that. Now I am gonna watch Iron Man and :sleeping: then.

But tomorrow, I think this could be finished (without return typehints - they will be solved after this is done - too many changes and merge conflicts).

vlastavesely:
@TomasVotruba Is the `ApiGen\Annotation\AnnotationSubscriber\ParamAnnotationSubscriber` used somewhere? It does not seem to me. Was it dead-end experiment?

TomasVotruba:
It's collected in DI here: https://github.com/ApiGen/ApiGen/blob/master/src/DependencyInjection/CompilerPass/CollectorCompilerPass.php#L38

And applied here, when `@param` annotation occurs
https://github.com/ApiGen/ApiGen/blob/master/packages/Annotation/src/AnnotationDecorator.php#L30

Based [on the code](https://github.com/ApiGen/ApiGen/blob/master/packages/Annotation/src/AnnotationSubscriber/ParamAnnotationSubscriber.php#L50), it should check annotation, where more classes are possible to pass.

```php
/**
 * @param Node|int
 */
```

vlastavesely:
The `ParamAnnotationSubscriber` seems to be ignored. I added exception throwing into `process()` but it did not fail. That is probably because of we are dealing with them directly loaded from `BetterReflection`. And with return annotations (when implemented) it will be the same.

What I am now struggling with is the centralized link generation - and this is in the subscribers. I am looking for a best way how to generate a link to reflection no matter whether it is return, param or see.

Now there are the subscribers on one hand and 

```
{foreach $property->getTypeHints() as $typehint}{$typehint|buildLinkIfReflectionFound}{sep}|{/sep}{/foreach}
```

on the other hand.

Could we somehow use existing subscribers or there will be a need for refactoring? Any ideas?

TomasVotruba:
To be honest, I'm lost in this PR. There to many changes and I don't use Apigen that much,so I would be shooting random.

Yet, feel free to remove the subscriber if it is not used. The mechanism can be still used on other annotations.

vlastavesely:
I would be best to start again from scratch :)

PR #977, opened 2017-08-10
[ecs] update to 2.2.6, make use of prepared configs

TomasVotruba:
Use [prepared configs](https://www.tomasvotruba.cz/blog/2017/08/07/7-new-features-in-easy-coding-standard-22/#2-prepared-configs) from ECS 2.2

This will be last bigger change for a long time.

### Changes 

- applied [Symfony checkers set](https://github.com/Symplify/EasyCodingStandard/blob/master/config/symfony-checkers.neon), apart few doc style rules
- added dot space checkers, `' . '` - 1 from each side
- last item in array has to end with comma `,`
- there is space after `! `, e.g. `if (! $autoloader)`
- partial uses [disabled](https://github.com/ApiGen/ApiGen/pull/977/files#diff-db71ac2c5cb10a341e624322fac72c1eR19), to make 1 way to import namespaces 

### Related

- disabled `SomeTrait`, it caused tests to fail: https://github.com/ApiGen/ApiGen/pull/977/files#diff-639c1b69f36cde9fa148db8015ee37e2
- fixed `getStartLine()` and `getEndLine()` in trait: https://github.com/ApiGen/ApiGen/pull/977/files#diff-8cf7e7f19d71a71b48d468ecce6d7826

@vlastavesely It would be great to remove or resolve all @todos, so they don't [polute config](https://github.com/ApiGen/ApiGen/pull/977/files#diff-db71ac2c5cb10a341e624322fac72c1eR88). This [might grow and cause everything](https://blog.codinghorror.com/the-broken-window-theory/) to be @todo and unfinished

### DIFF
--- a/bin/bootstrap.php
+++ b/bin/bootstrap.php
@@ -1,18 +1,18 @@
 <?php declare(strict_types=1);
 
-$rootDir = realpath(__DIR__.'/../');
+$rootDir = realpath(__DIR__ . '/../');
 
 // Autoloader for standalone install (installed via `composer create-project` or cloned locally)
-$autoloader = realpath($rootDir.'/vendor/autoload.php');
+$autoloader = realpath($rootDir . '/vendor/autoload.php');
 
-if (!$autoloader) {
+if (! $autoloader) {
     // Installed via `composer [global] require`.
-    $autoloader = realpath($rootDir.'/../../autoload.php');
+    $autoloader = realpath($rootDir . '/../../autoload.php');
 }
 
-if (!$autoloader) {
+if (! $autoloader) {
     throw new RuntimeException(
-        'ApiGen was unable to find its autoloader. '.
+        'ApiGen was unable to find its autoloader. ' .
         'Did you forget to run "composer update"?'
     );
 }
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "tracy/tracy": "^2.4",
         "phpunit/phpunit": "^6.0",
         "phpstan/phpstan": "^0.8",
-        "symplify/easy-coding-standard": "^2.2.3"
+        "symplify/easy-coding-standard": "^2.2"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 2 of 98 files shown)


TomasVotruba:
@vlastavesely Could I ask you for a review please? And merge if everything ok.

Going to meetup now.

vlastavesely:
Seems to be good. But personally I do not feel good about the comma after last element in array. That is required by some official standard?

I like the changes you made in `ecs`. Tomorrow, I will try it on my project how it works :+1:

----
Yeah, we should remove the `todo` annotations but we need to solve major things first. I think this is a minor one.

TomasVotruba:
> That is required by some official standard?

You can read more about it here: https://blog.madewithlove.be/post/code-style-options-for-cleaner-diffs/
See https://gist.github.com/hannesvdvreken/6ceaf8b553af057d8950/revisions vs. https://gist.github.com/hannesvdvreken/cadc3d8d6ff72e848dea/revisions

> I like the changes you made in ecs. Tomorrow, I will try it on my project how it works :+1:

Thank you :sun_with_face:

vlastavesely:
> You can read more about it here...

It seems to be reasonable. But from purely visual perspective, it seems a bad choice to me.

TomasVotruba:
Why do you think so?
It saved me lot of stress while adding new items.

vlastavesely:
Consider this:

```php
$a = [
  'a' => 'a',
  'b' => [
    'c' => 'c',
  ],
];
```

The last comma after the `]` is kinda strange here. I can't tell you why, it just feels so strange to me. Maybe because I've never seen it like this before...? But if you want this to be added, it is no problem.

TomasVotruba:
I had similar feeling like you. It is new and thus strange. But that's not reason not to try it.
Today I've read [Give it 5 minutes](https://signalvnoise.com/posts/3124-give-it-five-minutes), that helps me to try "strange" things, even though I disagree with them.

> But if you want this to be added, it is no problem.

Ok :)

vlastavesely:
So ready to merge?

TomasVotruba:
Yes :+1:

PR #987, opened 2017-08-26
[BetterReflection] improve parsing by Caching

TomasVotruba:
**This is attempt to make it faster, based on PR: https://github.com/Roave/BetterReflection/pull/312**

Basically just add cached PHP-Parser to BetterReflection SourceLocators.

The Parser was kinda slow, [see Blackfire](https://blackfire.io/profiles/d8df5224-ead3-4c18-a286-cb76d2ab4e9f/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=PhpParser%5CLexer%5CEmulative%3A%3AgetNextToken&callname=main()
).
 
image

### Related Changes

- move service configs to standalone files to their packages
- find a bug and fix it in `getOwnConstants()` of `ClassReflection` and `InterfaceReflection`


TomasVotruba:
Thanks @simara-svatopluk for live review. Much WTF found

TomasVotruba:
This improved performance from 23 s to 6 s on tested file.

Aiming on 1 s :)

TomasVotruba:
@vlastavesely Could I ask you for a review?

TomasVotruba:
Before this PR it takes 380 s to generate api for /src:https://travis-ci.org/ApiGen/ApiGen/builds/267478541#L821

Todo: check after merge

packages/BetterReflection/src/SourceLocator/ProjectVendorSourceLocatorFactory.php
```diff
@@ -0,0 +1,43 @@
...
+            return $composerSourceLocator;
+        }
+
+        return new AggregateSourceLocator;
+    }
+
+    /**
+     * @param string[] $directories
+     */
+    private function tryToFindProjectAutoload(array $directories): ?ComposerSourceLocator
```
simara-svatopluk: This method is still extremely difficult to read and understand for me. Suggestion, unfortunately not too better :-( : https://pastebin.com/YfcBhyyz

TomasVotruba: I see. Let's make it little bit better then, I found your code readable. Could you send PR please?

Just one note: Having `return dirname($directory) . self::STANDARD_AUTOLOAD_LOCATION;` inside another method would disable the traverse-up meaning of `dirname`. It has to be done before passing `$directory` as argument.

TomasVotruba:
Doctrine exceptions were not there before?
If not, could you provide minimal code that is causing that?

vlastavesely:
It worked before. Now it is failing, for `@ORM\Table(name="table")` needs to have space before the parentheses.

https://github.com/phpDocumentor/ReflectionDocBlock/blob/master/src/DocBlock/StandardTagFactory.php#L170

TomasVotruba:
Ok, I will look on that tomorrow.
I think some elements were not loaded before, e.g. getConstants() returned only own constants, see those 2 commits.

Maybe we could add some wrapper to resolve those annotation exceptions by ourselves.

TomasVotruba:
@vlastavesely I wonder, could you send PR with testsing Doctrine source to this PR?
It would drastically increase my motivation to finish this :)

vlastavesely:
You really think it should be solved within this PR? If you think so, I would create it tomorrow at the morning.

I had no time for ApiGen this week and it is not going to get better, at least until next week.

TomasVotruba:
@vlastavesely You're right, it's not really needed to fix that here.

PR #990, opened 2017-08-31
Improve ProjectVendorSourceLocatorFactory readability

simara-svatopluk:


TomasVotruba:
Thanks for the PR, nice job!

There is one thing that is not related to this PR, but to the BetterReflection
Could you also fix removed dependency?  

See: https://github.com/Roave/BetterReflection/commit/acb7899c623939d6de4cc0148db29bcfd9db487f#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780

```json
"require": {
    "zendframework/zend-code": "^3.1"
}
```

TomasVotruba:
Is ready to merge, I'll just wait for tests.

simara-svatopluk:
And I wondered why couple of tests didn't pass on my machine ;-)
You have a great checking process :+1:

TomasVotruba:
Mmm, this composer tests fails because we do shenaniganz with BetterReflection dependency. 
I think we can comment it out for now.

TomasVotruba:
Mmm, this composer tests fails because we do shenaniganz with BetterReflection dependency. 
I think we can comment it out for now.

TomasVotruba:
I'll handle it, thanks for the PR Svaťa!

PR #996, opened 2017-09-04
console: print info about parsing

### LINKED ISSUE #986: Missing progress bar or something like that
After running, Apigen several minutes shows nothing. It is not clear that it works and that I should wait.

vlastavesely:
It would close #986 

Prints out an information that it is doing something. After reflections parsed, the message is replaced with progressbar.

### DIFF
--- a/src/Application/ApiGenApplication.php
+++ b/src/Application/ApiGenApplication.php
@@ -60,12 +60,15 @@ final class ApiGenApplication
             DestinationOption::NAME => $runCommand->getDestination(),
         ]);
 
-        $this->output->write('Processing reflections (this may take a while)...');
+        $this->output->write('Parsing reflections (this may take a while)...');
         $this->parser->parseFilesAndDirectories($options[SourceOption::NAME]);
-        $this->output->write("\x0D\x1B[2K"); // clears the line
+        $this->output->writeln(' <info>done!</info>');
+        $this->output->writeln('Generating documentation...');
 
         $this->prepareDestination($options[DestinationOption::NAME], (bool) $options[OverwriteOption::NAME]);
         $this->generatorQueue->run();
+
+        $this->output->writeln('<info>Your documentation has been generated successfully!</info>');
     }
 
     private function prepareDestination(string $destination, bool $shouldOverwrite): void
--- a/src/Application/ApiGenApplication.php
+++ b/src/Application/ApiGenApplication.php
@@ -10,6 +10,7 @@ use ApiGen\ModularConfiguration\Option\OverwriteOption;
 use ApiGen\ModularConfiguration\Option\SourceOption;
 use ApiGen\Reflection\Parser\Parser;
 use ApiGen\Utils\FileSystem;
+use Symfony\Component\Console\Output\OutputInterface;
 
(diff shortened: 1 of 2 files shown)


TomasVotruba:
It could be good enough solution for now.

Just instead of passing dependency throught, you could require that in costructor.
See [Injecting dependencies just for passing them through](https://matthiasnoback.nl/2013/01/dependency-injection-smells/).

It's already registered as service, so it should be available: https://github.com/ApiGen/ApiGen/blob/bfc1d60a609d21a8edf103b009be62bee256dd96/src/config/services.yml#L40

src/Application/ApiGenApplication.php
```diff
@@ -52,7 +60,10 @@ public function runCommand(RunCommand $runCommand): void
             DestinationOption::NAME => $runCommand->getDestination(),
         ]);
 
+        $this->output->write('Processing reflections (this may take a while)...');
         $this->parser->parseFilesAndDirectories($options[SourceOption::NAME]);
+        $this->output->write("\x0D\x1B[2K"); // clears the line
```
TomasVotruba: Maybe it would be more clear to write "Parsing finished"
And bellow write "Generating documentation..."

vlastavesely: What about this?

screenshot-1

TomasVotruba:
Just few cs fixes, it's ready otherwise: https://travis-ci.org/ApiGen/ApiGen/builds/272898430#L756-L765

Btw, don't you use local `composer fs` before commit? It would fix this and it would be ready to merge now.

vlastavesely:
Usually I run `composer complete-check`. But sometimes I forget for another projects do not have it and  run just `vendor/bin/phpunit` 🐘

TomasVotruba:
Oh, I see.

Personally I'm too lazy to type more than 2 chars, so I just have bash shortcuts:

```bash
alias pu="vendor/bin/phpunit"
alias cs="composer check-cs"
alias fs="composer fix-cs"
alias ps="composer phpstan"
```

And I use 3 terminals, running `pu`, `cs` and `ps` to prevent waiting :rocket:

src/Application/ApiGenApplication.php
```diff
@@ -52,9 +60,15 @@ public function runCommand(RunCommand $runCommand): void
             DestinationOption::NAME => $runCommand->getDestination(),
         ]);
 
+        $this->output->write('Parsing reflections (this may take a while)...');
         $this->parser->parseFilesAndDirectories($options[SourceOption::NAME]);
+        $this->output->writeln(' <info>done!</info>');
```
TomasVotruba: Just a detail, in command below there is extra ` - `, so It might be added here to keep consistency.

```php
$this->output->writeln(' - <info>done!</info>');
```

vlastavesely: That was intentional. The three dots followed by the dash looked quite strange. What about this:

screenshot

TomasVotruba: I see. This looks good.

TomasVotruba:
Ready to merge

TomasVotruba:
Thank you!

PR #997, opened 2017-09-04
fixed Recoverable fatal error

vlastavesely:
Generating API for our company's system succeeded with one problem only. I am not really sure what class caused it but it was leading me to `trim` filter and after I removed it from `class.latte`, problem is gone. Strange is that it points to the project's dependencies and not ApiGen's:

```
Recoverable fatal error:  Object of class Latte\Runtime\FilterInfo could not be converted to string in /var/www/ivicom.cs/vendor/latte/latte/src/Latte/Runtime/Filters.php on line 600
```

What surprises me a bit is that API for Doctrine entities is generated without a fail. @TomasVotruba have you made any changes that could fix it somehow?

------

But still, ApiGen is really slow - it took one hour to generate complete API (233 entries).

### DIFF
--- a/packages/ThemeDefault/src/class.latte
+++ b/packages/ThemeDefault/src/class.latte
@@ -166,8 +166,7 @@
                 <tr>
                     <td>
                         {foreach $parentClass->getOwnProperties() as $parentClassProperty}
-                            <code>{block|trim}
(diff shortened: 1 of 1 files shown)


TomasVotruba:
I haven't changed anything since last PR. It might be related to BetterReflection updates.

>  I am not really sure what class caused it but it was leading me to trim filter and after I removed it from class.latte, problem is gone. Strange is that it points to the project's dependencies and not ApiGen's:

If you find what object is causing that anyd are you able to replicate, you could report that at [nette\latte](https://github.com/nette/latte/issues). We can merge it for now here.

PR #998, opened 2017-09-05
fix broken BetterReflection's behavior

vlastavesely:
I tested this:

```php
<?php

include __DIR__ . '/vendor/autoload.php';

use Roave\BetterReflection\BetterReflection;

$classInfo = (new BetterReflection())
    ->classReflector()
    ->reflect(Exception::class);

print_r(array_keys($classInfo->getInterfaces()));
```

it shown that problem is in BR.I updated dependency and two broken tests pass now. I will check the last also.


vlastavesely:
The relative path being used in `NamingHelper::nameToFilePath()` is used relatively to source parameter from config. In case of the broken test, the relative path is empty so the test should be fixed.

But I do not know whether you intended it or not. How it should behave ideally when there are multiple sources?

TomasVotruba:
Thanks a lot!

vlastavesely:
How do you mean that?

TomasVotruba:
I'm not sure anymore :D 

Where exactly is what bug? Here?

https://github.com/ApiGen/ApiGen/blob/9ada55bc2f3d8d7b7cbae64e17c4673035514770/src/Generator/FunctionGenerator.php#L81

tests/Generator/FunctionGeneratorTest.php
```diff
@@ -32,7 +32,7 @@ public function test(): void
         );
 
         $this->assertFileExists(
-            TEMP_DIR . '/source-function-Generator.Source.someFunction.php.html'
```
TomasVotruba: This should be kept. It should be unique per function.

These 2 files would be generated to 1 file (wrong):

```bash
/src/Generator/someFunction.php
/src/someFunction.php
```

vlastavesely: I have undone it but I am not sure about it.

If you provide `/tests/Generator/Source` as source path, relative path will be just `/SomeFunction.php`. For file `/tests/Generator/Source/AwesomeDir/SomeFunction.php`, the path would be different - it would be `/AwesomeDir/SomeFunction.php`.

Problem would be this:
```
apigen src src1 --destinaion docs
```
and having files `src/SomeFunction.php` and  `src1/SomeFunction.php` with function with same name.

TomasVotruba: Well that's an assumption that works for the most use-cases and has its limits.

How would you solve it?

TomasVotruba: On second though... I don't have time/capacity to solve this much further, so I'll be ok with any solution you pick best on your practise. **We can go with one you proposed.**

vlastavesely: Okay, I will extend the test and propose a final solution tomorrow.

vlastavesely:
I thought about it and updated the test. The only thing I am not sure about is how to handle cases where there are multiple sources contaning same files (`src/SomeFunction.php` and `another-src/SomeFunction.php`). For the sources may not have nothing in common, the only idea I have is to calculate a checksum of the source and prepend it before actual path so final path would be:

```
source-function-b9a23f-someFunction.php.html
source-function-e0692a-someFunction.php.html
```
@TomasVotruba what do you think about it?

TomasVotruba:
I think that will do. Failproof 👍

vlastavesely:
Oh, I am having a bad day :D

There is obvious error in the tests for the hash is generated for whole absolute path which is, of course, different on another machine. I will fix it later but you can see how I did mean it.

TomasVotruba:
@vlastavesely Oh, I'm sorry about that. I know how that feels.
I wish you good nerves - stay :muscle:

vlastavesely:
So, I am closing this also for it depended on your PR.

I am not having any time now, I am not sure when I will able to work on ApiGen but I would like to do all my best to make 5.0 release possible.

TomasVotruba:
> So, I am closing this also for it depended on your PR.

Sure.

> I am not having any time now, I am not sure when I will able to work on ApiGen but I would like to do all my best to make 5.0 release possible.

Take your time. It's also up to other contributors to make progress to ApiGen 5.

It helps me to take 1 PR at a time. It might be exhausting to manage multiple issues at once.

But the most: **enjoy it or don't do it** :+1:  It's open source after all :tulip:

Issue #935, opened 2017-07-07
PHP Notice:  Uninitialized string offset: 0 in '[...]ClassLoader.php on line 375


nelson6e65:
I'm getting this after #916 , as dependency (1), global dependency (2) and local clone (3):

:one: **`vendor/bin/apigen generate`**: 
`php_nml/vendor/composer/ClassLoader.php on line 375`

```
nelson6e65:~/workspace/php_nml (master) $ vendor/bin/apigen generate src --destination output/api
PHP Notice:  Uninitialized string offset: 0 in /home/ubuntu/workspace/php_nml/vendor/composer/ClassLoader.php on line 375

Notice: Uninitialized string offset: 0 in /home/ubuntu/workspace/php_nml/vendor/composer/ClassLoader.php on line 375

                                                                                               
  [Roave\BetterReflection\Reflector\Exception\IdentifierNotFound]                              
  Roave\BetterReflection\Reflection\ReflectionClass "" could not be found in the located sour  
  ce                                                                                           
                                                                                               

generate [--destination DESTINATION] [--] <source> (<source>)...
```

:two: **`apigen generate`**: 
`~/.composer/vendor/composer/ClassLoader.php on line 375`

```
nelson6e65:~/workspace/php_nml (master) $ apigen generate src --destination output/apiPHP Notice:  Uninitialized string offset: 0 in /home/ubuntu/.composer/vendor/composer/ClassLoader.php on line 375

Notice: Uninitialized string offset: 0 in /home/ubuntu/.composer/vendor/composer/ClassLoader.php on line 375

                                                                                               
  [Roave\BetterReflection\Reflector\Exception\IdentifierNotFound]                              
  Roave\BetterReflection\Reflection\ReflectionClass "" could not be found in the located sour  
  ce                                                                                           
                                                                                               

generate [--destination DESTINATION] [--] <source> (<source>)...
```

:three: **`../ApiGen/bin/apigen generate`**: 
`ApiGen/vendor/composer/ClassLoader.php on line 375`

```
nelson6e65:~/workspace/php_nml (master) $ ../ApiGen/bin/apigen generate src --destination output/api
PHP Notice:  Uninitialized string offset: 0 in /home/ubuntu/workspace/ApiGen/vendor/composer/ClassLoader.php on line 375

Notice: Uninitialized string offset: 0 in /home/ubuntu/workspace/ApiGen/vendor/composer/ClassLoader.php on line 375

                                                                                               
[...]

PR #1004, opened 2017-09-14
Fixes in @see annotations, added tests

vlastavesely:
I tried to locate source of problems from #935 but I found a few other problems with **see** annotation.

Currently I am stuck on problem that @dg had too here: 985. It is about the class named `Object`. I tested it and problem is not in directloading but in parent resolving. `BetterReflection` uses this: https://github.com/phpDocumentor/TypeResolver/blob/master/src/TypeResolver.php for resolving parent classes and thus `object` is resolved as PHP 7.2 typehint and not a class name.

For the moment, I do not know what solution would be the best? Any ideas?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     "bin": ["bin/apigen"],
     "require": {
         "php": "^7.1",
-        "roave/better-reflection": "dev-master#a402048",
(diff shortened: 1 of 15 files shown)


vlastavesely:
It looks like the new BR works with `Object` class correctly. Sadly, there are problems with typehints that need to be fixed :(

Issue #1007, opened 2017-09-27
apigen.yml is not identified by apigen on windows


vipinjn24:
Hi,

As in the instructions i have placed the apigen.yml file in the root of my project, but apigen ignores it.
Please see below the command used and the apigen.yml location.

apigen generate D:\Users\Vipin\Documents\PROJECT\prepare --destination D:\Users\Vipin\Documents\PROJECT\docs

The apigen.yml file is inside "D:\Users\Vipin\Documents\PROJECT\prepare" folder.

Please help me use it to work with custom themes.

contents of apigen.yml file:
parameters:
    visibilityLevels: [public, protected, private]
    annotationGroups: [todo, deprecated]
    title: "test"
    baseUrl: ""
    overwrite: true
    themeDirectory: "D:\Users\Vipin\Desktop\klarna-apigen-theme-master\src"

vipinjn24:
i am having this same issue in ubuntu also. do let me know if i am not placing the file in correct location.

mrtndwrd:
This is caused by #1025

vipinjn24:
@mrtndwrd Great you did the check on the issue 👍 Thanks

vipinjn24:
I got this error.

PHP Fatal error:  Uncaught Error: Cannot instantiate interface Roave\BetterReflection\SourceLocator\Ast\Locator in C:\Users\Vipin\AppData\Local\Temp\_apigen_kernel_cache\DependencyInjectionDevDebugProjectContainer.php:1446
Stack trace:
#0 D:\SERVER\WEBSERVER\PHP\composer\vendor\symfony\dependency-injection\Container.php(329): DependencyInjectionDevDebugProjectContainer->getRoave_BetterReflection_SourceLocator_Ast_LocatorService()
#1 C:\Users\Vipin\AppData\Local\Temp\_apigen_kernel_cache\DependencyInjectionDevDebugProjectContainer.php(413): Symfony\Component\DependencyInjection\Container->get('Roave\\BetterRef...')
#2 D:\SERVER\WEBSERVER\PHP\composer\vendor\symfony\dependency-injection\Container.php(329): DependencyInjectionDevDebugProjectContainer->getApiGen_BetterReflection_SourceLocator_DirectorySourceLocatorFactoryService()
[...]

PR #1026, opened 2017-11-20
Replace camelcase option names

### LINKED ISSUE #1025: apigen.yml ignores all options with upper case characters
(... except for baseUrl)

This is a rather surprising issue, because you already fixed it for one specific option: baseUrl. in the `resolveOptions` function in `Configuration/Configuration.php` you fix the `baseUrl` option as follows:

```php
        $baseUrlKeyLowered = strtolower(BaseUrlOption::NAME);                    
        if (isset($resolvedOptions[$baseUrlKeyLowered])) {                       
            $resolvedOptions[BaseUrlOption::NAME] = $resolvedOptions[$baseUrlKeyLowered];
        }  
```

The problem that you're fixing here, exists for all options that contain upper case characters (themeDirectory, visibilityLevels, annotationGroups). 

With this apigen.yml:

```yml
parameters:
  visibilityLevels: 
    - public
    - protected
  themeDirectory: "/vagrant/vendor/my-theme/src"
  title: "Title"
  baseUrl: "http://basurl.lan/docs"
  
```

I get the following `$options` array:

```php
Array
(
    [visibilitylevels] => Array
        (
            [0] => public
            [1] => protected
        )

    [themedirectory] => /vagrant/vendor/my-theme/src
    [title] => Title
    [baseurl] => http://baseurl.lan/docs
    [source] => Array
        (
            [0] => /vagrant/cosmos/protected/modules/cloud/models/Ip.php
        )

    [destination] => /vagrant/documentation/docs
    [annotationGroups] => Array
        (
        )

    [baseUrl] => 
    [overwrite] => 
    [themeDirectory] => /vagrant/vendor/apigen/apigen/packages/ModularConfiguration/src/Option/../../../../packages/ThemeDefault/src
    [visibilityLevels] => 768
)
```

All these other options are ignored. The problem, of course, is that somewhere they are added to the `$options` array with lower case characters, instead of the way they are actually filled in in the .yml file.

I'm currently trying to figure out *why* this happens. Does anyone already have an idea why these options lose their capitalisation once they have been read from the apigen.yml file?

mrtndwrd:
Use underscores instead of camelcase to fix problems caused by the fact that Yaml is case insensitive.

Closes #1025 and #1007

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ Associated element is internal, so ApiGen hides it.
 
 ## Themes
 
-To enable a custom theme just provide `themeDirectory` configuration option in your `apigen.yml`:
+To enable a custom theme just provide `theme_directory` configuration option in your `apigen.yml`:
 
 ```yaml
 parameters:
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Just look at [**ApiGen API**](https://apigen.github.io/ApiGen):
 
 ![ApiGen Preview](docs/preview.png)
 
-## Your Help is Needed to Finish 5.x Release  
+## Your Help is Needed to Finish 5.x Release
 
 :heart: We need your help to test new version of ApiGen.
 
@@ -78,10 +78,10 @@ the root of your project:
 
 ```yaml
 parameters:
-    visibilityLevels: [public, protected] # array
-    annotationGroups: [todo, deprecated] # array
+    visibility_levels: [public, protected] # array
+    annotation_groups: [todo, deprecated] # array
     title: "ApiGen Docs" # string
-    baseUrl: "http://apigen.org/api" # string
+    base_url: "http://apigen.org/api" # string
     overwrite: false # bool
 ```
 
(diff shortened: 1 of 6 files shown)


vipinjn24:
i have updated the code manually to this commit and still the options are not parsed and nothing happened.

vlastavesely:
@vipinjn24 So even with this changes it does not work for you? There must be something on your machine then. Maybe something Windows-related...?

vipinjn24:
@vlastavesely i checked with the code and found that the method 'registerContainerConfiguration' from 'AppKernel' class is never called where the configuration is picked.
Is this something windows related?

vipinjn24:
ok now i checked the behavior and found that `registerContainerConfiguration` is called the first time i called apigen but in this case it took my yml file from my current working directory and it now threw some error which i previously posted in https://github.com/ApiGen/ApiGen/issues/1007#issuecomment-346381397

Now i ran it after removing the yml file and it started running so i cancelled the job and now again i placed the yml file and after initiating the job found that `registerContainerConfiguration` in not being called no matter what i do!!!

vipinjn24:
ok why you are not merging this change in the master branch!!!

Its been a while since this pull is requested.

jankal:
@vipinjn24 Is this PR ready for merge (as in stable)?

vipinjn24:
@jankal yes this one can be merged.

PR #1027, opened 2017-11-20
use newer symfony to fix bug in parsing apigen.yml

mrtndwrd:
Here's an example of how you can test the themeDirectory configuration option. In this version, it will work, because I made it use Symfony 3.4 where this issue is fixed: symfony/symfony#23381 .

You might want to think about the following things though:

- Symfony 3.4 is not stable yet, so I had to reduce minimum-stability to make this work.
- The ApiGen documentation generation seems to use the same configuration file as the tests, so I had to change that file in order to be able to test the themeDirectory setting. In this case I just made a symlink to the normal theme directory, so it doesn't make a lot of difference...

Let me know if there's more I can do

### DIFF
--- a/apigen.yml
+++ b/apigen.yml
@@ -2,3 +2,4 @@ parameters:
     title: "ApiGen It-self"
     visibilityLevels: [public, protected]
     baseUrl: http://apigen.org
+    themeDirectory: tests/testThemeDir/src
--- a/composer.json
+++ b/composer.json
@@ -2,6 +2,7 @@
     "name": "apigen/apigen",
     "description": "PHP 7.1 source code API generator.",
     "license": "MIT",
+    "minimum-stability": "dev",
     "authors": [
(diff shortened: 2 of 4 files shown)


mrtndwrd:
This pull request was made as a reaction on the discussion in #1025

mrtndwrd:
Yeah, this is basically a proof of concept for the ideal fix...

What do you propose? At least this pr includes a failing test like you
asked for, so that should give you something to work with.

ADmad:
symfony 3.4 is stable now.

honeycombs2000:
symphony 3.4

jankal:
@mrtndwrd Please rais minimum-stability again. I will merge then.

JanTvrdik:
No longer relevant since https://github.com/ApiGen/ApiGen/releases/tag/v7.0.0-alpha

PR #1052, opened 2018-07-11
Better-Reflection dev-master#c87d856 is no more working #1046

dimabory:

Issue #1046, opened 2018-06-01
Better-Reflection dev-master#c87d856 is no more working


vipinjn24:
`"roave/better-reflection": "dev-master#c87d856"` is no more working and we have to use below.

`"roave/better-reflection": "dev-master"`

Please create a new PR to update the readme.md

simpsonw:
`"roave/better-reflection": "dev-master#c87d856"` and `"apigen/apigen": "dev-master"` don't appear to be working for me:

```
$ vendor/bin/apigen generate src/ --destination docs/apigen/
Parsing reflections (this may take a while)...PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Roave\BetterReflection\SourceLocator\Reflection\SourceStubber::addDocComment() must be an instance of PhpParser\BuilderAbstract, instance of PhpParser\Builder\Interface_ given, called in /var/www/html/project/vendor/roave/better-reflection/src/SourceLocator/Reflection/SourceStubber.php on line 63 and defined in /var/www/html/ual-payments/vendor/roave/better-reflection/src/SourceLocator/Reflection/SourceStubber.php:97
Stack trace:
#0 /var/www/html/project/vendor/roave/better-reflection/src/SourceLocator/Reflection/SourceStubber.php(63): Roave\BetterReflection\SourceLocator\Reflection\SourceStubber->addDocComment(Object(PhpParser\Builder\Interface_), Object(ReflectionClass))
#1 /var/www/html/project/vendor/roave/better-reflection/src/SourceLocator/Type/PhpInternalSourceLocator.php(53): Roave\BetterReflection\SourceLocator\Reflection\SourceStubber->__invoke(Object(ReflectionClass))
#2 /var/www/html/project/vendor/roave/better-reflection/src/SourceLocat in /var/www/html/ual-payments/vendor/roave/better-reflection/src/SourceLocator/Reflection/SourceStubber.php on line 97
```
Using `"roave/better-reflection": "dev-master"` gives me an exception when I try to run `composer update`:

```$ composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 1 update, 0 removals

                                                                 
  [InvalidArgumentException]                                     
  Package is not installed: roave/better-reflection-9999999-dev  
                                                                 

[...]

Issue #1020, opened 2017-11-02
Looking for a maintainer


vlastavesely:
There are many of us who would like to see version `5.0.0` of `ApiGen` to be released. Despite the fact that `5.0.0-RC5` has been released already, there is so much work to be done. Currently, there are some serious performance problems that must be resolved and some missing features that must be implemented, before we can release a final stable version.

This project, like every other open-source project, stands and falls with two things: collaboration and passion. It is really hard for an open source project to be successful when maintained by a single developer. This is true even more when the project is not a passion for its developers. If the project is not the most important thing in your development life, you should probably to give a chance to someone else.

When @TomasVotruba announced first release candidate, I started to help with implementing missing features and fixing issues I encountered with, in order to have `ApiGen` finished. But during the time, the performance issues have arisen that need deeper inspection and so much more work. I would like to have `ApiGen` finished to generate a sexy API reference for my project, but I cannot invest as much of effort as is needed here for I do have other projects that are priority in this time.

This is why we need your help. To be honest, without *your* help, this project will not be done. If it is important for you and you really want to have it finished, support it: fix an issue or implement some feature and send a PR. We will be happy to merge it.

But, what is most important, we are looking for a maintainer who would be able to put his or hers effort into massive optimization and making `ApiGen` competitive. Do you think that you right for the job, or do you know someone? Let us know!

jordanbrauer:
Are you only looking for a sole-maintainer? I just came across this tool the other day and I'd love to be apart of this project, however I am not wanting to head one right away that I am unfamiliar with.

vlastavesely:
Hi, any help would be appreciated. We do not expect that there would appear a new maintainer out of blue. Starting with a new and unfamiliar project is kind of hard so that is why I am here :) I have implemented a few features and fixed a few bugs so I have some knowledge of the code and therefore I can help you. And if we both would got stuck, I think that @TomasVotruba, who have implemented most things, would help also if it is needed.

Like I said, any help is appreciated here, even if it were just small baby-steps 👍

TomasVotruba:
Exactly as @vlastavesely said.
 I'm here if you need me :+1:

jordanbrauer:
Perfect 👍 I'd love to come on board then!

You mention the following in your OP,

>Despite the fact that `5.0.0-RC5` has been released already, there is so much work to be done. Currently, there are some serious performance problems that must be resolved and some missing features that must be implemented, before we can release a final stable version.

[...]

PR #1054, opened 2018-10-26
roave/better-reflection update. solves issues #1046 #1052  

theinrich88:
roave/better-reflection #1efdfe5 solved some type issues.
 #1020 I would be please to give a hand in this project


JanTvrdik:
No longer relevant since https://github.com/ApiGen/ApiGen/releases/tag/v7.0.0-alpha

PR #1101, opened 2024-03-04
Make Bootstrap compatible with Nette\Schema 1.3

AJenbo:
Both variants are there in 1.2 but only the later in 1.3:
https://github.com/nette/schema/blob/v1.2/src/Schema/Helpers.php

Projects current composer dependency is set to ^1.2 so this should align the code with that.


AJenbo:
Looks like you also wanted to update the dependency, your welcome I suppose:
https://github.com/ApiGen/ApiGen/commit/491ab6ae7ff95dbc20875f5118e836bff2155785

